PR #17, opened 2018-10-23
Remove no-dev option

shirshak55:
Because it is used in production and many people will use following command I think it is better to remove however warning can be made.

$ composer install --no-dev

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -18,7 +18,7 @@ Laravel Telescope is an elegant debug assistant for the Laravel framework. Teles
 
 You may use Composer to install Telescope into your Laravel project:
 
-    composer require laravel/telescope --dev
+    composer require laravel/telescope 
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -18,7 +18,7 @@ Laravel Telescope is an elegant debug assistant for the Laravel framework. Teles
 
 You may use Composer to install Telescope into your Laravel project:
 
-    composer require laravel/telescope --dev
+    composer require laravel/telescope
```
ntzm: There is a trailing space here :)

shirshak55: where?

martinbean: @shirshak55 After `laravel/telescope`.

driesvints:
Actually, we decided to explicitly add the `--dev` flag because you'd primarily use it in development. You can always install it as a normal requirement if you also want to run it in production.

PR #30, opened 2018-10-23
[1.0] update docblocks

browner12:
- `assignEntryPointFromBatch()` does not return anything
- `store()` and `destroy()` do not return anything

### DIFF
--- a/src/Http/Controllers/MonitoredTagController.php
+++ b/src/Http/Controllers/MonitoredTagController.php
@@ -42,7 +42,7 @@ class MonitoredTagController extends Controller
      * Begin monitoring the given tag.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return Response
+     * @return void
      */
     public function store(Request $request)
     {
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/MonitoredTagController.php
```diff
@@ -42,7 +42,7 @@ public function index()
      * Begin monitoring the given tag.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return Response
+     * @return void
      */
     public function store(Request $request)
```
quetzyg: Might as well add the return type

src/Http/Controllers/MonitoredTagController.php
```diff
@@ -53,7 +53,7 @@ public function store(Request $request)
      * Stop monitoring the given tag.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @return Response
+     * @return void
      */
     public function destroy(Request $request)
```
quetzyg: Same as above.

src/IncomingDumpEntry.php
```diff
@@ -18,7 +18,7 @@ public function isDump()
      * Assign entry point parameters from the given batch entries.
      *
      * @param  array  $batch
-     * @return $this
+     * @return void
      */
     public function assignEntryPointFromBatch(array $batch)
```
quetzyg: Same as above.

PR #38, opened 2018-10-23
Fixed argument error for EntryResult

cmdobueno:
Argument 6 in EntryResults needs to be an array and not null, so there was a check added to it.


themsaid:
The column isn't nullable in the DB, there's no way it would return null.

cmdobueno:
After further inspection, I believe this was caused by issues previous issues that were resolved in a different pull request. Somehow, I had a non-array. Basically from what I can tell, the column data became corrupt thus causing it to be invalid and no longer an array when passed to EntryResult. Upon clearing the tables, the error was gone. I feel there should still be a check though, as it is possible to no have an array.

PR #45, opened 2018-10-23
[1.0] Add required illuminate packages

Te7a-Houdini:


browner12:
If we're going to add this I would recommend first using tildes (`^`) for consistency, and secondly bumping them to `^5.7.7` since there is a method required that only exists in 5.7.7+.

Te7a-Houdini:
@browner12  thanks :+1:

X-Coder264:
`illuminate/support` should also be added.

Te7a-Houdini:
@X-Coder264  already required by  [illuminate/database](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Database/composer.json#L21)

X-Coder264:
And since 5.7.7+ is required I would switch back from caret to tilde (`^5.7.7` -> `~5.7.7`). The reason being that `^5.7.7` would allow 5.8, 5.9 etc to be used with this version of Telescope and since Laravel allows breaking changes in minor releases (it does not follow semantic versioning) it means that this version of Telescope cannot guarantee to work with all Laravel versions before Laravel 6 gets released.

X-Coder264:
It doesn't matter if `illuminate/database` requires it. What if `illuminate/database` drops that dependency tomorrow? :) `illuminate/support` is a clear dependency of this package and as such it should be added to composer.json in the same way as all other illuminate components.

Te7a-Houdini:
@X-Coder264  aha , i think you are right in this .  we should revert back to telda

for illuminate/support am not quite sure that it should be added explicitly  , as it's  required from other packages like ,  [illuminate/events](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Events/composer.json#L20) , [illuminate/http](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Http/composer.json#L18) , [illuminate/notifications](https://github.com/laravel/framework/blob/5.7/src/Illuminate/Notifications/composer.json#L25)

browner12:
I would agree with @X-Coder264 , this package requires it, so you shouldn't rely on another package to require it for you.

X-Coder264:
Yep, and `illuminate/contracts` should be added too.

taylorotwell:
Honestly for packages that are intended to always be included with the entire Laravel framework I don't even add this stuff.

PR #47, opened 2018-10-23
[1.0] Fix route caching

bonzai:
Right now you can't run `route:cache` because of closure.


X-Coder264:
You can use `Route::view()` to avoid having to create a controller.

PR #68, opened 2018-10-23
[1.0] Added `laravel/framework` to dependency

TBlindaruk:
- added "laravel/framework" : "^5.7.7", to `required`, since laravel should be with minimum version v5.7.7.

 So after this commit composer will update automaticly laravel to 5.7.7 or package will not be installed;

 Similar to https://github.com/laravel/telescope/pull/45

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,7 @@
     "require": {
         "php": "^7.1.3",
         "ext-json": "*",
+        "laravel/framework" : "~5.7.7",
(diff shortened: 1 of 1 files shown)


browner12:
as it was pointed out in the previous PR, this package is not guaranteed to work with 5.8, so the requirement should be `~5.7.7`.

also please make sure package requirements are in alphabetical order.

TBlindaruk:
fixed

Issue #55, opened 2018-10-23
Timezone is off if not set to UTC


viirre:
If I set another timezone other than `UTC` in my `config/app.php` file (eg `Europe/Stockholm`), all entries are reported as they happened with an offset (depending on the timezone defined). It appears as if you *must* have `UTC` in  `config/app.php` to get the correct time reported.

vesper8:
noticed that as well, my timezone is america/new york and things that happened just now are marked as having happened 4 hours ago

piotrgulbis:
confirming this. Have Europe/Vienna set in app.php and it's marking as happened two hours ago.

themsaid:
So in the database, what timezone are the entries stored in? The one set in config/app?

johanvanhelden:
Chiming in here. @themsaid the database entries are stored in the proper timezone.

Entries are stored as `2018-10-23 21:35:17` and that is also the current time. (Europe/Amsterdam)
However, the interface shows `2h ago`

viirre:
They are stored correctly as per the one specified in config/app

themsaid:
@virre I think that's an easy fix, will have to look into it tomorrow though :)

viirre:
Super! 👍

Mecit:
Are you sure this is actually fixed? I've just installed Telescope and my timezone in config/app.php is set to 'Europe/Istanbul' (that's GMT+3) and for a request made at **14:35**, the request details page says it was made at **5:35 PM (3h ago)**.

johanvanhelden:
Are you sure you installed the latest version? It has been released in 0.1.2.

themsaid:
@Mecit make sure you're on latest telescope and re-publish the assets.

ybaruchel:
@Mecit 
As @themsaid said, you'll have to republish the assets

Mecit:
Thanks @ybaruchel and @themsaid. Seems the package was still in v0.1.1 when I installed it, indeed. Updating and re-publishing the assets solved the issue.

hailwood:
Note to anyone who comes across this issue, republishes and still finds it doesn't work?
Try using the `--force` flag :)

yordivd:
I'm on version v1.2 and still facing this issue. Have republished and force republished.
The entry in my database table has the correct datetime.
for example 2019-10-25 09:44:00 is in my database entry
but in telescope it shows 2019-10-25 11:44:00s (5s ago)..
[...]

PR #69, opened 2018-10-23
[1.0] fix timezone format

ybaruchel:
Fix from [issue](https://github.com/laravel/telescope/issues/55)

- Added `moment-timezone`
- Exposed app's timezone to javascript through Telescope's "script variables". 
- Configured `moment-timezone` default timezone using timezone from "script variables"

** Made prod build instead of dev build - that's why there are so many deletions on the public folder.


themsaid:
@ybaruchel thanks man :)

PR #73, opened 2018-10-23
[1.0] Changed `DateTimeInterface` to `Carbon\Carbon` in `EntryResult`.

TBlindaruk:
- Changed `DateTimeInterface` to `Carbon\Carbon` in `EntryResult`, since in `jsonSerialize` we have used `$this->createdAt->toDateTimeString()`.

 toDateTimeString method not exist in `DateTimeInterface` but present on Carbon

### DIFF
--- a/src/EntryResult.php
+++ b/src/EntryResult.php
@@ -2,8 +2,8 @@
 
 namespace Laravel\Telescope;
 
+use Carbon\Carbon;
 use JsonSerializable;
-use DateTimeInterface;
 
 class EntryResult implements JsonSerializable
 {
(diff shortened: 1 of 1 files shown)


src/EntryResult.php
```diff
@@ -52,7 +52,7 @@ class EntryResult implements JsonSerializable
     /**
      * The datetime that the entry was recorded.
      *
-     * @var \DateTimeInterface
+     * @var Carbon
```
bonzai: `\Carbon\Carbon`

src/EntryResult.php
```diff
@@ -72,10 +72,10 @@ class EntryResult implements JsonSerializable
      * @param  string  $type
      * @param  string|null  $familyHash
      * @param  array  $content
-     * @param  \DateTimeInterface  $createdAt
+     * @param  Carbon  $createdAt
```
bonzai: `\Carbon\Carbon`

PR #77, opened 2018-10-23
Add notice about env

deleugpn:
I spent a while trying to figure out why Telescope wouldn't work on my projects. My company make use of `APP_ENV=testing` for local development and tests as we orchestrate the entire container suite through Docker compose and it's easier to track env vars through `.env.testing`


themsaid:
`testing` is the testing environment, I find it obvious that you should give a unique name to each environment.

PR #85, opened 2018-10-23
Name home route

enimiste:
This will make easy to define the Url in the menu using route('telescope::home') instead of writing url('/telescope'). The former will take in consideration the path defined in the config file without the need to edit the menu Url each time the configured path is edited.


bonzai:
Wrong branch.

enimiste:
Any explanation why it is closed ?

PR #92, opened 2018-10-24
(Update) README

HDVinnie:
- add php version requirement notice


webmasterdro:
It's works fine on PHP 7.1 (as in the composer).

PR #94, opened 2018-10-24
Fix bug

enimiste:
The url used in axios request is a relative url to the base url, so we don't need the "'/' +".
This caused an issue of receiving always the Http code 302. After diagnost the issue i found that the Route::post to "queries" for example never reached (I did a test using dd()). But when i removed the '/' from this file all works fine. (At this point you should disable the CSRF middleware to get thing working waiting for fixing the 419 HTTP code received if the Csrf middleware is enabled in Kernel.php).

### DIFF
--- a/resources/js/components/IndexScreen.vue
+++ b/resources/js/components/IndexScreen.vue
@@ -104,7 +104,7 @@
 
         methods: {
             loadEntries(after){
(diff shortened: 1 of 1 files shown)


themsaid:
your PR breaks all requests, will close for now until someone else reports the same issue so we can investigate further.

eduarguz:
IMHO, that should be reviewed again, this also happens in Horizon :(

Issue #91, opened 2018-10-24
Telescope creates migration in the wrong database


SerafimArts:
## Step To Reproduce

1. Run `php artisan telescope:install`.
2. Change `storage` => `database` => `connection` in `telescope.php` to another (`sqlite` as example).
3. Run `php artisan migrate`.

## Expected

Сreating a structure within the specified connection.

## Actual

Creating migrations on the primary connection.

SerafimArts:
![image](https://user-images.githubusercontent.com/2461257/47398190-a5d45b00-d73b-11e8-9b91-0dcdaa3ad5ef.png)

enimiste:
See PR #95

SerafimArts:
@enimiste thanks! Then it makes no sense to discuss it. Taylor don't care.

P.S. I recommend to use the [clockwork](https://github.com/itsgoingd/clockwork), it is more convenient and does not contain such problems.

PR #95, opened 2018-10-24
Fix bug related to the issue #91

enimiste:
The storage should be picked depending on the driver config not always 'database'


taylorotwell:
Incorrect. This binding is ONLY for the database driver. We wouldn't want to use any other connection.

enimiste:
Ok, but

Why there this driver config ?
Why the storage config is an array ?

This is the way Laravel config works. An array of possible drivers and the default one defined in a separted config.

PR #114, opened 2018-10-24
[1.0] Remove unused argument.

lucasmichot:
Remove unused argument, and ensure that `QueueController::show` is compatible with its parent method.


themsaid:
applied, thanks

PR #120, opened 2018-10-24
[1.0] Remove useless thrown docblocks

lucasmichot:
Remove useless thrown docblocks

### DIFF
--- a/src/ExtractProperties.php
+++ b/src/ExtractProperties.php
@@ -14,8 +14,6 @@ class ExtractProperties
      *
      * @param  mixed  $target
      * @return array
-     *
-     * @throws \ReflectionException
      */
     public static function from($target)
     {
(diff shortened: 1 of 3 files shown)


browner12:
why? these methods throw these exceptions, so if we're not going to catch them, we should alert people of this.

lucasmichot:
`\ReflectionClass::__construct` thrown an exception, but none of the methods for which this PR applied

lucasmichot:
At no moment we write `throw Exception...` in the code of `ExtractProperties@from()`

lucasmichot:
For laravel framework, see `\Illuminate\Filesystem\FilesystemAdapter::setVisibility`: it calls `\Illuminate\Filesystem\FilesystemAdapter::parseVisibility`

We don't set `@throws \InvalidArgumentException` in `setVisibility`...

PR #127, opened 2018-10-24
[1.0] Favor Yarn lock file

lucasmichot:
Favor Yarn lock file and stay consistant with other laravel projects.


browner12:
hmm.. if it's consistent with other projects that's good, but can we make a concerted effort to get rid of Yarn everywhere? It was a temporary solution until NPM got their stuff together. It's an extra dependency that we no longer need.

lucasmichot:
When using then Laravel installer, you make a `laravel new foobar`, it does not come with any `package-lock.json` file

I guess you can propose it to Taylor to always have bother version available (yarn.lock and package-lock.json)

browner12:
no, definitely don't want both versions installed. I would say we just have `package-lock.json` or get rid of both of them.

IMO we should start trying to wean people off of Yarn because it's an extra dependency that we no longer need.

PR #130, opened 2018-10-24
Moving the migrations outside the `src` folder

arcanedev-maroc:


themsaid:
I think it's better to keep it there so all storage related stuff in one place.

PR #132, opened 2018-10-24
Add the `provides()` method to the service provider

arcanedev-maroc:


taylorotwell:
Why?

taylorotwell:
This method is only useful when a service provider is deferred.

arcanedev-maroc:
OK :+1:

PR #143, opened 2018-10-24
[1.0] Use the database connection from the config

### LINKED ISSUE #91: Telescope creates migration in the wrong database

simonellensohn:
The migrations are executed on the default DB connection instead of the one provided in the telescope config file.

Fixes #91

### DIFF
--- a/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
+++ b/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
@@ -7,11 +7,11 @@ use Illuminate\Database\Migrations\Migration;
 class CreateTelescopeEntriesTable extends Migration
 {
     /**
-     * The database connection.
+     * The database schema.
      *
-     * @var string
+     * @var Schema
      */
-    protected $connection;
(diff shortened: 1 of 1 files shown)


src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
```diff
@@ -6,14 +6,31 @@
...
+
     /**
      * Run the migrations.
      *
      * @return void
      */
     public function up()
     {
-        Schema::create('telescope_entries', function (Blueprint $table) {
+        Schema::connection($this->connection)->create('telescope_entries', function (Blueprint $table) {
```
lucasmichot: Maybe simply create a variable for `Schema::connection($this->connection)` in the constructor

PR #144, opened 2018-10-24
[1.0] Remove useless elseif statements.

lucasmichot:
Remove useless `elseif` statements.


browner12:
how are these "useless"? it is simply a different coding style.

is there a performance implication?

is there precedent for style consistency? if that's the case, I'm 👍 , otherwise I don't think these changes are necessary.

themsaid:
we'll keep it that way for now.

Issue #129, opened 2018-10-24
X-Request-Id Header


dennisoderwald:
It would be useful if Telescope had the option to enable a middleware that sets the ID of the request previously in `X-Request-Id` header.

This makes it easier to analyze errors, especially in productive environments with a large number of requests per second.

themsaid:
Can you explain more?

dennisoderwald:
The telescope stores each operation (requests etc.) under a UUID. The UUID would set the middleware as response header, so that the UUID can be used to find the request faster. Normally you use the X-Request-Id header for this.

themsaid:
We have no plans to support this at the moment. Might be an idea for the future.

PR #152, opened 2018-10-24
[1.0] Set Telescope Request ID

lucasmichot:
Set Telescope Request ID.
This PR is for https://github.com/laravel/telescope/issues/129 by @dennisoderwald

---

The header name is called `X-Telescope-Request-Id` as other popular debugging/logging SAAS already use `X-Request-Id` and we want to avoid conflicts.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -2,6 +2,7 @@
 
 use Laravel\Telescope\Watchers;
 use Laravel\Telescope\Http\Middleware\Authorize;
+use Laravel\Telescope\Http\Middleware\SetTelescopeRequestId;
 
 return [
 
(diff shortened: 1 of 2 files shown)


themsaid:
I don't think that's what he meant, I don't see why setting this header for Telescope requests is useful.

dennisoderwald:
@themsaid Not for requests to `/telescope` but for all requests that are logged to assign a specific request faster.

The implementation of @lucasmichot randomly generates a UUID which probably does not match the UUID of the Telescope Entry.

barryvdh:
You probably want to move the UUID generation for the request entry before this middleware, then inject that header in the response. (Or the batch ID?)

PR #161, opened 2018-10-24
General cleanup/improvements

X-Coder264:
Some general cleanup and improvements:

- the concrete `Container` contract implementation also implements `ArrayAccess` to resolve dependencies, but that isn't on the `Container` contract and as such stuff like `$app['events']` would break on any other implementation
- I type-hinted stuff as much as possible to avoid issues like https://github.com/laravel/telescope/commit/fa6919367e9d0af7bbac05c0b3fa89707239010d#diff-deecd2e5f86cd9955088f4026ca3d452 (and to enable better IDE auto-completion) and in doing so I found some instances of calls being made to a contract implementation which doesn't have that method (`getKey` was called on an `\Illuminate\Contracts\Auth\Authenticatable` implementation and `makeWith` was called on an `\Illuminate\Contracts\Container\Container` implementation)
- some controller docblocks said they were returning a `Response` instance or just "mixed", but they actually returned a `View` instance - I switched them all to actually return a proper `Response` object

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -52,11 +52,12 @@ Telescope exposes a dashboard at `/telescope`. By default, you will only be able
  *
  * This gate determines who can access Telescope in non-local environments.
  *
+ * @param  \Illuminate\Contracts\Auth\Access\Gate  $gate
  * @return void
  */
-protected function gate()
+protected function gate(Gate $gate)
 {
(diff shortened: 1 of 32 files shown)


taylorotwell:
A lot of this stuff isn't needed and some if it is actively bad... such as type-hinting Closure when other invokables could be used.

PR #162, opened 2018-10-24
General cleanup/improvements

X-Coder264:
Resubmit of https://github.com/laravel/telescope/pull/161

I've removed all the controller changes and the Closure type-hint. If you still think some of the remaining stuff isn't needed please say what exactly so that I can remove it from the PR.


browner12:
that's a lot of code/files to review. If you can make the changes smaller, might be more likely to be accepted.  It's hard to discern the intent and benefit of each change when there are so many.

src/Telescope.php
```diff
@@ -123,20 +126,20 @@ protected static function runningApprovedArtisanCommand($app)
                 'horizon',
                 'horizon:work',
                 'horizon:supervisor',
-            ], config('telescope.ignoreCommands', []))
+            ], $app->make(Repository::class)->get('telescope.ignoreCommands', []))
```
lucasmichot: Why not simply `$app['config']->get(` ?

src/Telescope.php
```diff
@@ -123,20 +126,20 @@ protected static function runningApprovedArtisanCommand($app)
...
+     * @param  \Illuminate\Contracts\Foundation\Application  $app
      * @return bool
      */
-    protected static function handlingNonTelescopeRequest($app)
+    protected static function handlingNonTelescopeRequest(Application $app)
     {
-        return ! $app->runningInConsole() && ! $app['request']->is(
-            config('telescope.path').'*',
+        return ! $app->runningInConsole() && ! $app->make(Request::class)->is(
+            $app->make(Repository::class)->get('telescope.path').'*',
```
lucasmichot: same

src/Telescope.php
X-Coder264: @lucasmichot I've explained that in https://github.com/laravel/telescope/pull/161

lucasmichot:
Some are nice such as type-hinting variables in closure some are more personal preferences like getting rid of the facades...

X-Coder264:
@lucasmichot I don't know why was there a need for "..." and the thumbs down reaction as all changes are like you said "nice" or "personal preferences" aka "neutral" - there isn't a single thing that is "worse" in this PR than before (plus it actually fixes some "broken" stuff).

@browner12 I had a bit of spare time to go through the entire package today and do this, if this PR doesn't get accepted I'll submit smaller PRs sometime in the future when I'll have some free time again like today.

Issue #116, opened 2018-10-24
[feature request] back trace support for queries


johanvanhelden:
I'd be really useful to see what part of the code caused a query. Like this:

![screenshot from 2018-10-24 11-29-27](https://user-images.githubusercontent.com/19389981/47421063-17ce9380-d780-11e8-9888-64bb85a35f0d.png)

That way you know where to look if something goes wrong or something needs to be optimized.

PR #164, opened 2018-10-25
Show backtrace for queries

cmorbitzer:
This change shows the backtrace for queries as requested in #116.

screen shot 2018-10-25 at 1 12 53 am

screen shot 2018-10-25 at 1 13 02 am

Some of the PHP code is loosely derived from https://github.com/barryvdh/laravel-debugbar/blob/v3.2.0/src/DataCollector/QueryCollector.php
I can add an MIT notice if you believe it would be required.

### DIFF
--- a/resources/js/components/PreviewScreen.vue
+++ b/resources/js/components/PreviewScreen.vue
@@ -76,6 +76,14 @@
         methods: {
             prepareEntry() {
                 document.title = this.title + " - Telescope";
+                this.ready = false;
+
+                let unwatch = this.$watch('ready', (newVal, oldVal) => {
+                    if (newVal) {
+                        this.$emit('ready');
+                        unwatch();
+                    }
+                });
 
                 this.loadEntry((response) => {
                     this.entry = response.data.entry;
--- a/resources/js/screens/queries/preview.vue
+++ b/resources/js/screens/queries/preview.vue
@@ -83,9 +83,7 @@
                 </ul>
 
                 <div>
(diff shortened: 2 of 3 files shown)


barryvdh:
> I can add an MIT notice if you believe it would be required.
No need for that, for me atleast.

Perhaps strip the `base_path()` from the final URL, to make it only show the relevant part?

cmorbitzer:
> Perhaps strip the base_path() from the final URL, to make it only show the relevant part?

I wanted to do that but changed my mind once I saw that the full path is shown for exceptions, which is where I borrowed the UI for this. Also, since Telescope can support multiple applications, seeing the full path might help distinguish which application server a query was executed from.

I'm not opposed to it, though.

johanvanhelden:
Thanks for the PR @cmorbitzer !

I agree with @barryvdh - I'd prefer to only see the relevant part.

taylorotwell:
Doing a full debug_backtrace() for every single query in the application is going to be a performance drain.

johanvanhelden:
What about making it optional via a config option that is disabled by default? That way, people have a choice.

cmorbitzer:
@taylorotwell Agreed. Would you accept a PR that made it optional as @johanvanhelden suggests?

During development or troubleshooting, the added performance cost of this feature is well worth the saved time and effort.

johanvanhelden:
Precisely. I'm hoping this gets added. That would make it a complete Debugbar replacement for me. Right now I still need to have both installed for query debugging.

barryvdh:
A simpler version is added here: https://github.com/laravel/telescope/pull/179

PR #165, opened 2018-10-25
Add env for TELESCOPE_DB_CONNECTION.

yajra:
- This will allow telescope to use a separate DB connection aside from the main.
- This will also allow support for third party DB drivers that is not officially supported by Laravel. For my case, Oracle DB using my package [Laravel-OCI8](https://github.com/yajra/laravel-oci8).

Thanks!

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -22,7 +22,7 @@ return [
 
     'storage' => [
         'database' => [
-            'connection' => env('DB_CONNECTION', 'mysql'),
+            'connection' => env('TELESCOPE_DB_CONNECTION', env('DB_CONNECTION', 'mysql')),
         ],
     ],
 


shirshak55:
Hmm i think we should have dedicated telescope config file where all such envs are listed and we can cache configs :)

emielmolenaar:
What's wrong with altering the published config file that lives in your applications' config/ directory?

shirshak55:
@emielmolenaar  sorry i didn't know it was added  just 2 day ago

https://github.com/laravel/telescope/commit/4ca19521bd98de20f97a75b80a0fb1a2b899b69f

yajra:
> What's wrong with altering the published config file that lives in your applications' config/ directory?

Not much, It will just save us some time in editing incase you re-publish the config.

yajra:
Also, it seems like publishing of [telescope config is optional](https://github.com/laravel/telescope/blob/1.0/src/TelescopeServiceProvider.php#L100-L102). I do prefer to use the default config provided by the vendor as much as possible to avoid any future config changes conflict :)

emielmolenaar:
@shirshak55 No problem!

@yajra On the other hand, publishing your own configuration ensures that configuration for your application, preventing accidental configuration changes caused by package updates. This is a consideration on has to make for each application I guess :)

yajra:
>  preventing accidental configuration changes caused by package updates

Good point.

emielmolenaar:
As I said... This is a consideration one has to make for each application I guess :)

taylorotwell:
You can change this in your own application.

PR #175, opened 2018-10-25
Add "telescope.enabled" option

tillkruss:
Disabling Telescope in production is quite annoying. This hidden config option that defaults to `true` would allow everything to be loaded (migrations, commands, views), but tracking can be disabled easily.

We might be able to use this in other places as well.

### DIFF
--- a/src/TelescopeServiceProvider.php
+++ b/src/TelescopeServiceProvider.php
@@ -22,8 +22,10 @@ class TelescopeServiceProvider extends ServiceProvider
         $this->registerMigrations();
         $this->registerPublishing();
 
-        Telescope::start($this->app);
-        Telescope::listenForStorageOpportunities($this->app);
+        if (config('telescope.enabled', true)) {
+            Telescope::start($this->app);
+            Telescope::listenForStorageOpportunities($this->app);
+        }
 
         $this->loadViewsFrom(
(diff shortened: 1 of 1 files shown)


browner12:
could you update the config stub, too?

themsaid:
We already have options to disable each watcher, you can disable them all if you want.

tillkruss:
I'm currently doing this:

```php
    'watchers' => env('TELESCOPE', false) ? [
        Watchers\CacheWatcher::class => env('TELESCOPE_CACHE_WATCHER', true),
        Watchers\CommandWatcher::class => env('TELESCOPE_COMMAND_WATCHER', true),

        // ... snip

        Watchers\RedisWatcher::class => env('TELESCOPE_REDIS_WATCHER', true),
        Watchers\RequestWatcher::class => env('TELESCOPE_REQUEST_WATCHER', true),
        Watchers\ScheduleWatcher::class => env('TELESCOPE_SCHEDULE_WATCHER', true),
    ] : [],
```
Because I don't want to disable 15 watchers using environment variables.

I do want to load the `Laravel\Telescope\TelescopeServiceProvider`, but without calling start and registering a bunch of stuff listeners.

tillkruss:
Alright, I'll go with your `extra.laravel.dont-discover` suggestion.

PR #188, opened 2018-10-26
Add OPTIONS as a method

DanielKeeble:
Add OPTIONS as a method so that it gets styled as a badge.

screenshot 2018-10-26 at 13 39 28


fh32000:
@DanKeeble @taylorotwell 
How i can ignore OPTIONS from report ?

Issue #186, opened 2018-10-26
TelescopServiceProvider should have the right namespace


m-lotze:
After `php artisan telescope:install` TelscopeServiceProvider has the namespace App\Providers. 
This should be changed to the namespace set for the app.

sifex:
Added #197 to Fix

themsaid:
fixed.

PR #197, opened 2018-10-27
Fixes Custom Namespacing Issues (#186)

sifex:
Tested against a new Laravel Installation, all works fine

* Essentially carries over the `file_put_contents` method to the Service Provider as well.
* Replaces the namespace using the DetectsApplicationNamespace trait


sifex:
**Improvements to be made:**
* Don't rely on the fact that someone has the EventService Provider

PR #198, opened 2018-10-27
[1.0] Modify using echo json_encode to @json directive

Te7a-Houdini:


taylorotwell:
I don't think it's currently broken.

PR #199, opened 2018-10-27
[1.0] Add telescopeApiPath to the root component

Te7a-Houdini:

PR #203, opened 2018-10-27
[1.0] Resubmitting #199 with explanation (Add telescopeApiPath to the root component)

Te7a-Houdini:
this **PR** , extracts the repeated ```'/' + Telescope.path + '/telescope-api/' ``` across axios requests & href values to a **data property** in the **root Vue instance** called ```telescopeApiPath```

 i thought in case the **Telescope window object** changed the **path** property name or in case the ```/telescope-api``` changed , so there will be only one abstracted place called **telescopeApiPath** where our changes can be reflected to different axios calls or href values  from single place

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -49,7 +49,9 @@ new Vue({
                 confirmationCancel: null,
             },
 
-            autoLoadsNewEntries: localStorage.autoLoadsNewEntries === '1'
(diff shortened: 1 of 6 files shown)


themsaid:
Hey,

Sorry but I don't see we're changing this as it'll be a major breaking change so it's ok to have it like this for now.

thanks

PR #216, opened 2018-10-28
Move moontoast/math to dev dependencies.

lucasmichot:
Move `moontoast/math` to dev dependencies.


taylorotwell:
I think it's needed as a regular dependency.

PR #217, opened 2018-10-28
add data variable to be used in template

okaufmann:
This will fix the following error on the mail preview screen:

image


themsaid:
Thank you, solved this in a different way :)

PR #223, opened 2018-10-28
Added missing ignoreCommands to config

paras-malhotra:


taylorotwell:
I will add it and follow the standard convention for comments.

PR #226, opened 2018-10-28
Enhancement - app identifier

Donny5300:
This PR will allow you to add a "app_identifier" to your application. The identifier will display which server made the action. The `Dumps` page is excluded with displaying the app identifier.

schermafbeelding 2018-10-29 om 00 54 48

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -7,6 +7,11 @@ return [
 
     'path' => 'telescope',
 
+    'app_identifier' => [
+        'name' => env('TELESCOPE_APP_IDENTIFIER', 'application'),
+        'enabled' => true
+    ],
+
     /*
     |--------------------------------------------------------------------------
     | Telescope Storage Driver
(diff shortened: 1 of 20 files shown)


themsaid:
Telescope isn't meant to collect logs from multiple applications, we already record the hostname in every entry so you know which server sent these actions, but not different apps.

deleugpn:
@themsaid given the new rise of microservice, I think this would be a pretty good feature. I understand the implementation seem a bit cumbersome and might need some work, but having multiple microservices (multiple laravel applications) writing to the same Telescope database would be useful since the Software Solution provided is a single one.

Donny5300:
@themsaid console commands doesn't use URL's so it would be hard to track on which server something is running. Same thing with Jobs, Events and all other background tasks.

taylorotwell:
It's a pretty decent idea but I don't like it taking up a column in every single table. I would rather have it feel more intentionally baked in like some kind of dropdown that lets you switch between available applications, etc.

Donny5300:
@taylorotwell how about your dropdown idea next to the top 2 buttons with available servers? Or next to the "search tag" input?  If a user selects a server, the column will disapear from the view. If a user doesnt select a server, it will display the column. The bad thing about removing the column from the view is that you need to switch servers everytime if you want to look through your exceptions (or any other item).

PR #231, opened 2018-10-29
Use uuid for uuid

Shhu:
Hi,

I'm using this pattern on my apps and it broke telescope : 
`Route::pattern('id', '\d+');`

Since telescope use uuid wich should be matched by : 
`Route::pattern('uuid', '[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}');`

### DIFF
--- a/src/Http/routes.php
+++ b/src/Http/routes.php
@@ -2,60 +2,60 @@
 
 // Mail entries...
 Route::post('/telescope-api/mail', 'MailController@index');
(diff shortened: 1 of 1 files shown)


deleugpn:
We currently don't have tests for any of these routes. Did you make sure that Telescope still works after this change?

PR #233, opened 2018-10-29
added favicon

powolnymarcel:


deleugpn:
I don't think this is in the right folder?

driesvints:
This should go in the `public` directory and be properly linked in https://github.com/laravel/telescope/blob/ee75daea25a1b6547e44bc90ae50bf9cffbf96b8/resources/views/layout.blade.php

PR #249, opened 2018-10-29
Added ClearCommand to clear all Telescope entries

### LINKED ISSUE #20: Anyway to clear log and disable log for sometime ?
There is `telescope:prune` which i though will prune the logs but it is not working .

And there is someway to disable logging untill we specify start logging or like that. There is button that will prevent from loading but it still catches logs.

Thanks Themsaid again :)

paras-malhotra:
This PR adds a `ClearCommand` console command class to clear all Telescope entries. It's useful to clear all Telescope entries, perform an action and then view fresh entries to debug applications. This command will enable clearing all entries.

Fixes #20. Although the issue also mentions disabling logs for some time / pausing logs, this PR only addresses clearing logs.

### DIFF
--- a/src/Console/ClearCommand.php
+++ b/src/Console/ClearCommand.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace Laravel\Telescope\Console;
+
+use Illuminate\Console\Command;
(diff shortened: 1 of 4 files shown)


taylorotwell:
Merged.

PR #255, opened 2018-10-30
[Coverage] Test Case for Clear Command

deleugpn:
While writing the Test Case for ClearCommand I noticed that the type-hint interface does not have the `clear()` method.


themsaid:
Please avoid changing the code, the PR description says it's only test coverage related.

PR #256, opened 2018-10-30
Added option to avoid pruning monitored tags

### LINKED ISSUE #111: [Suggestion] Option to prevent monitored tags from being pruned?
I'm trying to keep an eye on which queries are being tagged `slow` but because my app is processing a lot of requests my monitored tags rapidly get pruned.

Would it benefit others as well to have a switch next to each monitored tag to disabled pruning of those tagged entries?

I believe this would also be useful to tag a specific user, switch the tag to 'prevent pruning', then have them perform a specific action.

paras-malhotra:
This PR adds a config option to avoid pruning monitored tags. This is especially helpful in production to avoid rapid pruning of monitored tags when processing a large number of requests. Fixes #111 

This PR contains the following 2 breaking changes and is therefore targeted at the 1.0 branch:

1. The `config/telescope.php` limit option is now under the `pruning` key so that all configs related to pruning are grouped together.
2. Skipping monitored tags is enabled by default and this may change behaviour of pruning

I would be happy to resubmit without the above breaking changes if you'd like to target a minor release.


themsaid:
Please submit with only minimal changes.

paras-malhotra:
Sure @themsaid. Can you elaborate on which changes I should avoid. Are you referring to the config / breaking changes? If you're referring to the extra `EntryTagModel`, I can remove it but the query will be a little more complex since I won't be able to use `whereDoesntHave`.

themsaid:
I think we should wait before we go into implementing this for now, some people might want to configure it per tag so if you take the path you provide here it might make it harder for us in the future to introduce another breaking change.

paras-malhotra:
@themsaid, seems to me that not pruning monitored tags should be a pretty common use case (esp in production apps). So, I do think we shouldn't wait for a future PR to merge this.

Specific tag pruning seems like a less common use case. I could possibly also add the option to prune specific tags if you'd like. 

The config could be something like:
```
'dont_prune' => 'monitored_tags', // one of monitored_tags, specific_tags or prune_all
'exclude_tags' => [], // for specific tags not to prune (only when the specific_tags option is selected
```

Let me know if that sounds like a more generic solution. I would be happy to work on it if you agree or alternatively, open to any other suggestions.

themsaid:
Check the changes in: https://github.com/laravel/telescope/commit/f097b7e6d6c319ccee3750f2a6f178d1c352b596

It should stop pruning entries that belong to monitored tags.

paras-malhotra:
This is amazing and with minimal changes!

Some observations:

1. I think we should add the same implementation in the `prune` method of the DatabaseEntries. E.g. the `PruneCommand` relies on the same method.
2. I think the implementation breaks the `limit` feature. As an example, lets say there are 200 entries, 100 of which have monitored tags and lets say the limit is 50. The `pruneEntries` first whereNotIn will exclude the first 50 entries (of which say 10 are monitored tags). The next whereNotIn will exclude all monitored tags. So, we would be left with 100 monitored tags + 40 non-monitored tags. Whereas I think we should be left with 100 monitored tags + 50 non-monitored tags. Essentially the second `whereNotIn` condition should also apply in the inner query of the first `whereNotIn` condition IMO.
3. This always avoids pruning monitored tag entries. Something we need to consider is whether everyone wants this feature. I do think this is the 99% use case though, so I think it should be fine.

paras-malhotra:
Never mind, I guess pruning has been removed as a feature.

PR #257, opened 2018-10-30
Fix: when laravel is in subfolder

jjhamkuijpers:
This pull request makes it possible to let telescope still work if the project is in a subfolder.


themsaid:
this is not correct.

PR #261, opened 2018-10-30
[1.0] Fix PruneCommand contract

deleugpn:
The PruneCommand relies on the `EntriesRepository`, which does not have the `prune` method. Moreover, if the developer tries to swap the `PrunableRepository` in the container, it wouldn't work because of the wrong type-hint.

This Pull Request changes the contract being used by the PruneCommand to rely on the correct contract.


taylorotwell:
Can you resubmit without conflicts.

PR #268, opened 2018-10-30
Remove trimming of entries

themsaid:
If people want to remove older entries they should use the prune console command or implement their own scheduled job.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -28,21 +28,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Record Pruning
-    |--------------------------------------------------------------------------
-    |
-    | This configuration options determines how many Telescope records of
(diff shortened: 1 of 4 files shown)


deleugpn:
Running services behind a VPC on Amazon means I don't have an easy way of running Console commands since I cannot `ssh` into AWS Fargate and I also don't plan on having a container running 24/7 just to run Laravel Schedule. The simple pruning system was really good for GDPR compliance.
Do you plan on giving us anything similar again?

themsaid:
@deleugpn you can run your own code in a terminable middleware, that's how trimming used to work.

tillkruss:
Is this mentioned in the docs?

paras-malhotra:
I've created a package for intelligent pruning: https://github.com/insenseanalytics/laravel-telescope-pruning

Comments/suggestions would be most appreciated!

PR #290, opened 2018-10-31
[Coverage] Add Job Watcher Coverage

deleugpn:


deleugpn:
The problem is caused by this change https://github.com/laravel/framework/pull/26250
Since it's a static variable, it carries on accumulating lots and lots of `createPayloadUsing` between tests and once it gets trigger, it creates 30+ rows worth of values. It's passing on `setup=lowest` but failing on `latest`.
I'm working on figuring out where the fix should be.

tjmartin69:
@deleugpn Many thanks I have `Queue::createPayloadUsing(null)` on my test teardowns already.

crynobone:
@deleugpn Should be able to send a PR to <https://github.com/orchestral/testbench-core> for this changes, but technically this should goes into Laravel Framework first.

PR #295, opened 2018-11-01
[1.0] Add decoding uri params for get requests in requests index screen

### LINKED ISSUE #140: Add GET params to Requests Index screen
I think it would be best if the index screen already printed out the query string, because sometimes it is difficult to differentiate between two requests with the same path, but different query string.

So basically if the request is of type "GET" then append the "payload" property to the output string.

Te7a-Houdini:
this **PR** resolves #140  and decodes the payload in requests index screen to be like this screenshot

screenshot from 2018-11-01 08-43-55


themsaid:
we'll start capturing the full URL starting next release.

PR #298, opened 2018-11-01
Snake case config option

freekmurze:
All config options in Laravel are snake_cased, there's no reason why this one should be camelCased. I left support for the old casing to ensure backwards compatibility.

Imho this option should also be added to the default config file, but I didn't want to invest time now to come up with a comment of exactly three lines, each one three characters shorter than the other. 😀

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -123,7 +123,7 @@ class Telescope
                 'horizon',
                 'horizon:work',
                 'horizon:supervisor',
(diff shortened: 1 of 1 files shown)


deleugpn:
The broken test has nothing to do with this PR. It was introduced by https://github.com/laravel/telescope/commit/15960082a6dbe27780cc6f540fe102cef09ce8f6. I'll fix it in another PR.

PR #302, opened 2018-11-01
Updated dependencies

mrmajik45:
You can look at the commit history. I'm not going to copy it down using phone.


taylorotwell:
Most of these are covered by the existing dependency requirements?

mrmajik45:
But I just did a fresh fork on this?

PR #310, opened 2018-11-02
Implement shouldIgnore method for other watchers

guney:
@taylorotwell @themsaid  How should we proceed with tests? Should I change the environment setup to something like:

```php
$app->get('config')->set('telescope.watchers', [
    RequestWatcher::class => [
        'enabled' => true,
        'ignore' => [
             'something/*',
        ],
    ],
]);
```

or do you have some other and better suggestion?

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -49,7 +49,8 @@ return [
     |
     | The following array lists the "watchers" that will be registered with
     | Telescope. The watchers gather the application's profile data when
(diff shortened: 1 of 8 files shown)


deleugpn:
I can check the Test Suite later.

themsaid:
For example you can ignore all `OPTIONS` requests, or all events except for a few, using `Telescope::filter` is much more powerful.

PR #311, opened 2018-11-02
Remove adding TelescopeServiceProvider to app.php in installCommand because package-auto-discovery

ggcasuso:
Telescope already implement package-auto-discovery so isn't necessary to add TelescopeServiceProvider to app.php.

In my case I add to one project and install locally, added telescope files to gitignore (because I only want it in local env), but the installation process modify my app.php so I have to remove manually the register of the provider and if I forget it, it breaks my app.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -54,18 +54,6 @@ class InstallCommand extends Command
     {
         $namespace = str_replace_last('\\', '', $this->getAppNamespace());
 
-        $appConfig = file_get_contents(config_path('app.php'));
-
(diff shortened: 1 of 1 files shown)


taylorotwell:
This is the app Telescope service provider, not the vendor Telescope one.

ggcasuso:
Ok but the telescope service provider still copying to providers folder, why telescope app needs to add  to autoload service providers list in config/app.php if already have package-auto-discovery?

Sorry but I'm working with laravel for a short time and maybe I'm missing something.

paras-malhotra:
@ggcasuso Auto-discovery only works with packages (refer [the docs](https://laravel.com/docs/5.7/packages#package-discovery) for more details). The Telescope app service provider is part of your application code, so it is not auto-discovered.

Issue #265, opened 2018-10-30
[Feature Request] Duration of request


deviouspk:
Hi,
Would it be possible to log and display the time it took to execute a request? This could be very useful.

deviouspk:
On #313 taylor said:
![image](https://user-images.githubusercontent.com/7100315/47967349-a97db100-e05c-11e8-830f-c8ebbb202879.png)

I would love to have it on the index though. There's plenty of place to display it. And i really don't want to click on each request to check the execution time of it.
Can we have a discussion about this?

deleugpn:
I think it would be better to have a configurable index in such a way that each one can choose what to see in their index given each object's context.
That would free the burden of Telescope having to please too many variables.

PR #313, opened 2018-11-02
record request duration

vjolenz:
Hi,

I think it would be good to record requests' durations and some already requested on #265.


taylorotwell:
I wouldn't include it on the index. Maybe only the preview.

PR #319, opened 2018-11-03
Add ability to filter batches of entries

### LINKED ISSUE #318: [Feature Request] No way to filter by batch
To the best of my knowledge, currently there is no way to filter entries by batch. For example, say we need to record all entries for an app lifecycle that result in an exception (e.g. the queries fired, the request, etc.).

`Telescope::filter` only works for filtering individual entries but does not allow filtering entries at a batch level. Is there a way to do this in the existing package? If not, should we add something like `Telescope::filterBatch`? I would be happy to send a PR if this is useful. It's definitely a feature I'd like to have in all my apps using Telescope.

paras-malhotra:
Currently there is no way to filter batches of entries. For example, say we need to record all entries for an app lifecycle that result in an exception (e.g. the queries fired, the request, etc.).

This PR fixes #318 and adds the ability to filter batches like so:

```php
Telescope::filterBatch(function ($entries) {
	return $entries->contains(function($entry) {
		return $entry->isReportableException();
	});
}
```

The above code enables us to record all entries for any app lifecycle that results in a reportable exception. If this is merged, I think we should also change the `TelescopeServiceProvider.stub` to add this as an example.

On a side note, I saw many methods in `Telescope.php` that `return new static`. I don't know if this serves any purpose. Perhaps we should delete that and just have them as `void` methods since I can't find any non-static member or function in the `Telescope` class. Currently, I have followed the same pattern in my PR for consistency.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -28,6 +28,13 @@ class Telescope
     public static $filterUsing = [];
 
     /**
+     * The callbacks that filter the entries that should be recorded.
+     *
+     * @var array
+     */
+    public static $filterBatchUsing = [];
+
+    /**
      * The callback that adds tags to the record.
      *
      * @var \Closure
@@ -423,6 +430,19 @@ class Telescope
     }
 
     /**
+     * Set the callback that filters the batches that should be recorded.
+     *
+     * @param  \Closure  $callback
+     * @return static
+     */
+    public static function filterBatch(Closure $callback)
+    {
+        static::$filterBatchUsing[] = $callback;
+
+        return new static;
+    }
+
+    /**
      * Set the callback that adds tags to the record.
      *
      * @param  \Closure  $callback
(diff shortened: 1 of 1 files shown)


deleugpn:
Returning new static means we can chain calls.

paras-malhotra:
@deleugpn how would that work on static methods?

paras-malhotra:
Never mind, I got it. Static methods can be called from an object context. Didn't know that. Thanks!

paras-malhotra:
@taylorotwell batch here means an application lifecycle. If you'd like me to rename the methods, I'd be happy to do so.

taylorotwell:
Hmm, yeah, I understand the usefulness of filtering out an entire request / CLI / whatever... but I do think the method naming needs a little work.

paras-malhotra:
@taylorotwell we could consider renaming the app cycle level filter to `filterAll`. Alternatively, we could rename the existing `filter` to `shouldRecordEntry` or `filterRecording` (individual entry level) and the app cycle level filter to `shouldStoreEntries` or `filterStoring` (collection of entries).

Let me know your thoughts!

PR #322, opened 2018-11-03
Fix RedisWatcher by enabling events

### LINKED ISSUE #321: RedisWatcher doesn't record entries as the event is not dispatched
By default, events are disabled in `RedisManager` due to this commit: https://github.com/laravel/framework/commit/b14cfc9966572cf7b61e05a433947895cfddef2a

Unless these events are enabled `RedisWatcher` would not work as it listens to the `CommandExecuted` event. That's why it's broken right now.

paras-malhotra:
RedisWatcher currently does not record Redis entries since Redis events are disabled by default. This PR enables events to allow RedisWatcher to record Redis entries. Fixes #321

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,12 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+services:
+  - redis-server
+
 before_install:
   - phpenv config-rm xdebug.ini || true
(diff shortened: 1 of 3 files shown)


themsaid:
We can't enforce enabling events on each app, you have to do it explicitly in your App Service provider if you want.

PR #327, opened 2018-11-04
Replace empty request paths by a simple slash

jcqtc:
Pic related, avoid empty paths

capture


taylorotwell:
Can you submit a way to do this without using str_split? Seems like a odd way to check for this.

PR #328, opened 2018-11-04
Add request metadata: controller, action, middleware, duration

### LINKED ISSUE #265: [Feature Request] Duration of request

### LINKED ISSUE #314: [Feature Request] Track the Route/Controller/Trait/Views used in a response
Similar to the functionality of the Routes and Views tabs in Laravel Debugbar, I'd love to see support for tracking which route matched for a request as well as the views called in a request and the Endpoint Method that returned the request. Tracking these values would allow me to replace Debugbar/Clockwork with Telescope.
screenshot from 2018-11-02 12-41-32

paras-malhotra:
This PR adds the following request metadata to the detail screen:

1. Controller and action
2. Middleware
3. Request duration (in ms)

Fixes #314 and fixes #265. Here's how the screen looks like:

screen shot 2018-11-04 at 11 00 51 pm

### DIFF
--- a/src/Watchers/RequestWatcher.php
+++ b/src/Watchers/RequestWatcher.php
@@ -8,6 +8,7 @@ use Laravel\Telescope\Telescope;
 use Laravel\Telescope\IncomingEntry;
 use Symfony\Component\HttpFoundation\Response;
 use Illuminate\Foundation\Http\Events\RequestHandled;
+use Illuminate\Routing\Router;
 
 class RequestWatcher extends Watcher
 {
@@ -33,11 +34,14 @@ class RequestWatcher extends Watcher
         Telescope::recordRequest(IncomingEntry::make([
             'uri' => str_replace($event->request->root(), '', $event->request->fullUrl()),
             'method' => $event->request->method(),
+            'controller_action' => optional($event->request->route())->getActionName(),
+            'middleware' => optional($event->request->route())->gatherMiddleware(),
             'headers' => $this->headers($event->request->headers->all()),
             'payload' => $this->payload($event->request->all()),
             'session' => $this->payload($this->sessionVariables($event->request)),
             'response_status' => $event->response->getStatusCode(),
             'response' => $this->response($event->response),
+            'duration' => defined('LARAVEL_START') ? floor((microtime(true) - LARAVEL_START) * 1000) : null,
         ]));
     }
 
--- a/src/Watchers/RequestWatcher.php
+++ b/src/Watchers/RequestWatcher.php
@@ -35,7 +35,7 @@ class RequestWatcher extends Watcher
             'uri' => str_replace($event->request->root(), '', $event->request->fullUrl()),
             'method' => $event->request->method(),
             'controller_action' => optional($event->request->route())->getActionName(),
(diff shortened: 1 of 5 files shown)


deviouspk:
I would also love to see the duration of the request in the index page.

jcqtc:
Maybe we could hide the controller's namespace if it is `App\Http\Controllers` ?

paras-malhotra:
Personally, I think the duration should be only on the detail page and the complete namespace should be visible. You can discuss suggestions with @themsaid and @taylorotwell. I'll be happy to make changes accordingly.

jcqtc:
I find that including the whole namespace is less "eye catchy", but removing the application's namespace might be confusing...

It would be nice to have the namespace greyed out (but still clearly visible)

deleugpn:
I don't use the default Laravel namespace. My applications are modular and usually have controllers like `App\Modules\Revenue\Controllers`, so keeping the namespace is good for me.

src/Watchers/RequestWatcher.php
```diff
@@ -33,11 +34,14 @@ public function recordRequest(RequestHandled $event)
...
             'uri' => str_replace($event->request->root(), '', $event->request->fullUrl()),
             'method' => $event->request->method(),
+            'controller_action' => optional($event->request->route())->getActionName(),
+            'middleware' => optional($event->request->route())->gatherMiddleware() ?? [],
             'headers' => $this->headers($event->request->headers->all()),
             'payload' => $this->payload($event->request->all()),
             'session' => $this->payload($this->sessionVariables($event->request)),
             'response_status' => $event->response->getStatusCode(),
             'response' => $this->response($event->response),
+            'duration' => defined('LARAVEL_START') ? floor((microtime(true) - LARAVEL_START) * 1000) : null,
```
deleugpn: I don't think we need to check if it has been defined?

paras-malhotra: @deleugpn the tests fail if not defined since it isn't defined in the testbench. Also, this variable is a part of the `laravel/laravel` repo not the `laravel/framework`, so is susceptible to changes. That's why I've added this check.

deleugpn: I just proposed https://github.com/laravel/laravel/pull/4845. If accepted, we would be able to clean this up a bit and even test it properly.

paras-malhotra: Appreciate your thoughts although I don't understand completely. The change you've proposed is in the `laravel/laravel` repo. This doesn't impact the `orchestra/testbench` library so we'll still need a check to see if it's defined / resolvable from the container. Perhaps, I may be missing something. 

Also, I see the `laravel/laravel` repo as a starter that people are free to change. So I'm not sure if we can be confident that this would even be defined in apps using the Laravel framework.

laurencei: Even if the PR is accepted - you cant change the code above. The changes you've proposed in the other PR only apply to new projects. Most will be using the old code, and therefore you need to continue to check if the constant is defined or not.

PR #338, opened 2018-11-06
Use JSON blade helper.

lucasmichot:
Use`@json` blade helper.

I don't know what is the purpose of the `array_merge`, but I guess it can be removed.


bonzai:
Already proposed and rejected in #198.

lucasmichot:
Indeed

PR #340, opened 2018-11-06
Use `$this->assert`

lucasmichot:
Use `$this->assert` in place of `self::assert`.
There are no differences in term of performances or assertion.
But the non-static call is more consistent with the tests found in Laravel repositories.


deleugpn:
These functions are declared `static` on PHPUnit Framework. Ref: https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Assert.php#L1397

lucasmichot:
I know.
They are called non-statically in all the other Laravel repositories.

PR #365, opened 2018-11-10
[1.0] Extract some logic from the layout into the controller.

lucasmichot:
Extract some logic from the layout into the controller.

### DIFF
--- a/src/Http/Controllers/HomeController.php
+++ b/src/Http/Controllers/HomeController.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Telescope\Http\Controllers;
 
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/HomeController.php
```diff
@@ -13,6 +14,9 @@ class HomeController extends Controller
      */
     public function index()
     {
-        return view('telescope::layout');
+        return response()->view('telescope::layout', [
```
Te7a-Houdini: is there a reason to use ```response()->view()```  instead of ```view()``` directly ?

PR #366, opened 2018-11-10
[1.0] Extract all fixtures to their own namespace.

lucasmichot:
Extract all fixtures to their own namespace.

### DIFF
--- a/tests/Fixtures/Authenticated.php
+++ b/tests/Fixtures/Authenticated.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Laravel\Telescope\Tests\Fixtures;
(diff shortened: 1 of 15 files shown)


paras-malhotra:
Personally, I think the tests are much more readable if the classes stay in the test class. This is consistent with the Laravel framework tests as well.

lucasmichot:
Hm, I see your point (even if I have the opposite opinion), but it also depends on the project:
See https://github.com/laravel/socialite/tree/3.0/tests/Fixtures

deleugpn:
I also prefer the fixtures in the test file like the Framework code.

PR #367, opened 2018-11-10
[1.0] No need to use config() helper.

lucasmichot:
No need to use  `config()` helper as we already have access to `$this->app`


paras-malhotra:
You mean config() helper :)

lucasmichot:
True @paras-malhotra ! 😅

taylorotwell:
It's shorter to use config helper though?

PR #368, opened 2018-11-11
[1.0] Improve performance: dont register watchers on ignored commands / routes

paras-malhotra:
Currently, the watchers are registered and listen to events (and store entries in memory) even on ignored commands or routes. This PR improves Telescope performance by only registering watchers & events on approved commands/routes.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -103,7 +103,9 @@ class Telescope
      */
     public static function start($app)
     {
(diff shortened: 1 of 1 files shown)


paras-malhotra:
@themsaid thanks, I'll try and re-submit without this bug. Will possibly add regression tests to confirm.

PR #369, opened 2018-11-11
[1.0] Hide security critical fields to avoid session highjacking and csrf exploits

paras-malhotra:
This PR hides the csrf token and cookie in the session, request and headers. The cookie is hidden entirely with this PR (primarily to avoid displaying the session cookie). Although we could have added something like `static::$hiddenCookies`, in my opinion this is best left to the app user.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -62,6 +62,9 @@ class Telescope
      */
     public static $hiddenRequestHeaders = [
         'authorization',
+        'cookie',
+        'x-csrf-token',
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm, I think I would maybe stub these into the Telescope service provider instead. During local development I don't see any reason to hide them and in fact showing them could be helpful at times in that scenario.

paras-malhotra:
Makes sense, I'll send across a separate PR after adding it to the stub with an environment check.

PR #370, opened 2018-11-11
[1.0] Add ability to ignore routes with better performance

paras-malhotra:
This PR adds the ability to add ignored routes. Although this could be done with `Telescope::filter`, adding it to config is better performance due to #368.

I haven't changed the config file due to the comment conventions (both `ignore_commands` and `ignore_routes` are currently missing in `config/telescope.php`). I can take a go at it as a separate PR once (if) this is merged.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -105,7 +105,7 @@ class Telescope
     {
         if ($app->runningUnitTests() ||
             ! (static::runningApprovedArtisanCommand($app) ||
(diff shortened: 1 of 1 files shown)


taylorotwell:
The naming of your configuration option may make people assume you mean named routes (I assumed that) but you actually are talking about URIs.

paras-malhotra:
Got it, I'll update the PR to `ignore_uris` instead.

PR #373, opened 2018-11-12
[1.0] Add config options for ignored paths and commands

paras-malhotra:
As a follow up to #370 and #223, I've taken a stab at adding config options for ignored paths and ignored commands. I've followed the comment convention but I understand that Taylor is very picky on config (and rightly so) :grin: so feel free to close if it doesn't make sense.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -44,6 +44,35 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Ignored Paths
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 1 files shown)


vesper8:
very useful addition!

in laravel/telescope/src/Telescope.php there's this oddity with both ignoreCommands and ignore_commands

```
], config('telescope.ignoreCommands', []), config('telescope.ignore_commands', []))
```

Also in the codebase I found no use of the string "ignore_paths" anywhere

Issue #377, opened 2018-11-13
FileNotFoundException gets thrown when uploading files


johanvanhelden:
In our application we have multiple file uploads. This code works flawlessly, but as soon as we activate Telescope, we get a `FileNotFoundException`.

This is the full stack trace:

```
[2018-11-13 13:54:28] local.ERROR: The file "/tmp/phpm0ijmq" does not exist {"userId":10,"email":"redacted@redacted.nl","exception":"[object] (Symfony\\Component\\HttpFoundation\\File\\Exception\\FileNotFoundException(code: 0): The file \"/tmp/phpm0ijmq\" does not exist at /var/www/projects/redacted/vendor/symfony/http-foundation/File/File.php:37)
[stacktrace]
#0 /var/www/projects/redacted/vendor/symfony/http-foundation/File/UploadedFile.php(77): Symfony\\Component\\HttpFoundation\\File\\File->__construct('/tmp/phpm0ijmq', true)
#1 /var/www/projects/redacted/vendor/laravel/framework/src/Illuminate/Http/UploadedFile.php(120): Symfony\\Component\\HttpFoundation\\File\\UploadedFile->__construct('/tmp/phpm0ijmq', 'php-uninstall.t...', 'text/plain', 0, false)
#2 /var/www/projects/redacted/vendor/laravel/framework/src/Illuminate/Http/Concerns/InteractsWithInput.php(332): Illuminate\\Http\\UploadedFile::createFromBase(Object(Symfony\\Component\\HttpFoundation\\File\\UploadedFile))
#3 [internal function]: Illuminate\\Http\\Request->Illuminate\\Http\\Concerns\\{closure}(Object(Symfony\\Component\\HttpFoundation\\File\\UploadedFile))
#4 /var/www/projects/redacted/vendor/laravel/framework/src/Illuminate/Http/Concerns/InteractsWithInput.php(333): array_map(Object(Closure), Array)
#5 /var/www/projects/redacted/vendor/laravel/framework/src/Illuminate/Http/Concerns/InteractsWithInput.php(331): Illuminate\\Http\\Request->convertUploadedFiles(Array)
#6 [internal function]: Illuminate\\Http\\Request->Illuminate\\Http\\Concerns\\{closure}(Array)
#7 /var/www/projects/redacted/vendor/laravel/framework/src/Illuminate/Http/Concerns/InteractsWithInput.php(333): array_map(Object(Closure), Array)
[...]

PR #379, opened 2018-11-13
Format file uploads while recording a request

themsaid:
This PR fixes https://github.com/laravel/telescope/issues/377

We'll iterate over file uploads and display a placeholder instead of trying to use the actual file.

screen shot 2018-11-14 at 4 05 40 pm

### DIFF
--- a/src/Watchers/RequestWatcher.php
+++ b/src/Watchers/RequestWatcher.php
@@ -40,7 +40,7 @@ class RequestWatcher extends Watcher
             'controller_action' => optional($event->request->route())->getActionName(),
             'middleware' => optional($event->request->route())->gatherMiddleware() ?? [],
             'headers' => $this->headers($event->request->headers->all()),
-            'payload' => $this->payload($event->request->all()),
+            'payload' => $this->payload($this->input($event->request)),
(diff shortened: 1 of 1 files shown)


src/Watchers/RequestWatcher.php
```diff
@@ -97,6 +97,23 @@ private function sessionVariables(Request $request)
...
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return array
+     */
+    private function input(Request $request)
+    {
+        $files = $request->files->all();
+
+        array_walk_recursive($files, function (&$value) {
+            $value = 'File detected by Telescope';
```
Te7a-Houdini: i think if we used something like this code ,we will be able to show more info related to the uploaded file , instead of using string placeholder

```php
array_walk_recursive($files, function (&$value) {
   $value = ['name' => $value->getClientOriginalName(), 'size' => $value->getClientSize()];
});
```

taylorotwell: @themsaid Might be a good idea.

themsaid: @Te7a-Houdini great idea!

PR #380, opened 2018-11-13
[1.0] Fix docblocks.

lucasmichot:
Fix docblocks.

### DIFF
--- a/src/Http/Controllers/RecordingController.php
+++ b/src/Http/Controllers/RecordingController.php
@@ -28,7 +28,7 @@ class RecordingController extends Controller
     /**
      * Toggle recording.
      *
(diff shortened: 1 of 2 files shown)


src/Http/Controllers/RecordingController.php
```diff
@@ -28,7 +28,7 @@ public function __construct(CacheRepository $cache)
     /**
      * Toggle recording.
      *
-     * @return \Illuminate\Http\JsonResponse
+     * @return int|null
```
paras-malhotra: shouldn't this be `bool|void`? `forget` returns bool and if the key wasn't found, the function doesn't return anything

PR #397, opened 2018-11-16
Allow for setting age of records for pruning

denjaland:
Pruning was set to 24 hours in code.
I'd like to be able to set it to another value so we have a longer history available (7 days for instance).

thank you.


ludo237:
It would be nicer to allow ENV file to set it

themsaid:
Makes more sense to have it as an optional argument in the console command.

PR #398, opened 2018-11-16
[1.0] Provide Migrate command for telescope-only

deleugpn:
When running Telescope on a telescope-only dedicated database, even if we do `php artisan migrate --database=telescope` Laravel will still try to run all migrations (telescope and app) into the telescope-only database. Not only that's a bit wasteful, it also creates confusion when looking at the database. The developer might think he's deleting the telescope database when in fact he's deleting the app's database.

With `php artisan telescope:migrate`, only the Telescope migrations will be executed in the target connection. So if the target connection is a dedicated for telescope, it will only have the telescope tables.

### DIFF
--- a/src/Console/MigrateTelescopeCommand.php
+++ b/src/Console/MigrateTelescopeCommand.php
@@ -2,8 +2,8 @@
 
 namespace App\Console\Commands;
 
-use Illuminate\Config\Repository;
 use Illuminate\Console\Command;
+use Illuminate\Config\Repository;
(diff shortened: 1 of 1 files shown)


themsaid:
People can run `php artisan migrate --database={connection} --path=vendor/laravel/telescope/src/Storage/migrations`

deleugpn:
Thanks. Will test that to confirm it accomplishes my desired output.

PR #399, opened 2018-11-16
Display entry's hostname

warrenca:
I am working on a multiple projects and they are all using the same database. Currently, there's no way to know where the request came from, unless you'll dig inside the request details. I thought it's going to be useful to show the hostnames directly on the main requests page, differentiating them with colors. I've puta globe button beside the "Path" on the header to toggle the display of the hostname.

Hostnames are shown

Hostname are hidden but each request is still distinguished by color

FYI I have no vue experience prior to this, so suggestions are highly appreciated.


themsaid:
Sorry but no plans to add this at the moment.

PR #400, opened 2018-11-16
Add query explains for MySQL driver

MehediDracula:
[Explain](https://dev.mysql.com/doc/refman/8.0/en/explain.html) can be useful to [improve query performance](https://dev.mysql.com/doc/workbench/en/wb-tutorial-visual-explain-dbt3.html) while developing an app.

A new `Explains` table will be shown under the `Query` card in the query details page if the query has any explain data. Which type of queries will gather explain information, can be controlled from the config.

telescope-explain-queries

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -87,6 +87,10 @@ return [
         Watchers\QueryWatcher::class => [
             'enabled' => env('TELESCOPE_QUERY_WATCHER', true),
             'slow' => 100,
+            'explain' => [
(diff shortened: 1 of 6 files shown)


taylorotwell:
No plans to add this right now.

MehediDracula:
Is there any particular reason for not adding this? Since this is disabled by default, people won't even notice anything unless they enable it.

I think it could be useful for some people who used Laravel Debugbar with explain option enabled.

PR #410, opened 2018-11-18
Fix container contract violation

X-Coder264:
The `makeWith` method is not on the contract and it is just an alias for the `make` method which is on the contract.


paras-malhotra:
@X-Coder264 Telescope does not refer to the contract in the package. It always refers to the concrete implementation so I don't see it as a violation.

PR #415, opened 2018-11-20
Allow telescope to be enabled in testing

specialtactics:
It's pretty useful to be able to use telescope during automated testing (say to debug some specific tests). However at the moment, there is no way at all to have telescope enabled during automated test runs and no way to configure this.

This allows the developer to enable telescope in a testing env with a simple .env switch.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -39,6 +39,8 @@ return [
 
     'enabled' => env('TELESCOPE_ENABLED', true),
 
+    'enabled_in_testing' => env('TELESCOPE_ENABLED_IN_TESTING', false),
+
     /*
     |--------------------------------------------------------------------------
     | Telescope Route Middleware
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm, really I would rather change how Telescope works internally to just always rely on TELESCOPE_ENABLED and do away with that check to see if we're running tests. Probably TELESCOPE_ENABLED = false environment variable should be added to a project's `phpunit.xml` file.

specialtactics:
Hey @taylorotwell 

At the moment TELESCOPE_ENABLED won't have any effect in phpunit.xml file. Can you suggest how we might do this better? Happy to play around with a different approach to improve this aspect of telescope.

lk77:
Hello,

this is still an issue, 
when using laravel dusk i can't use telescope because APP_ENV is on testing during the dusk tests

driesvints:
@lk77 add this to your phpunit.xml: https://github.com/laravel/laravel/blob/8.x/phpunit.xml#L29

lk77:
yes, it's already disabled on dusk side.
APP_ENV should be local for telescope to run

PR #416, opened 2018-11-20
Fix listening for JobFailed event

blaxxi:


paras-malhotra:
Thanks @blaxxi. Fixes #416

PR #421, opened 2018-11-21
Update layout.blade.php

patrickadvance:
Added a Home Button for easier navigation to the home page from telescope dashboard


themsaid:
There's no home screen for Telescope :)

patrickadvance:
The button takes you back to your main domain. Eg. exampe.com/telescope/requests to example.com

PR #422, opened 2018-11-21
Automatically turns on Telescope::night() from 6pm to 6am.

patrickadvance:
This turns on the night mode automatically during night hours.


themsaid:
Thanks but better leave it for users to decide. I use night mode all day and night for example :D

PR #426, opened 2018-11-22
Fix storing session errors field

virp:
ViewErrorBag can't be serialized to json and stored as empty object '{}', it's becouse $basg (in ViewErrorBag) array of the view error bags is protected.


themsaid:
Please open an issue with steps to replicate. So far I haven't seen any troubles.

PR #432, opened 2018-11-23
Allow Telescope routes to be "domain-based"

jankremlacek:
Such as Nova, for example, it would be nice to allow to configure Telescope to serve routes only on a specific domain.

Taken Nova as a "best practice", the PR does not include pre-defined config in the configuration file. Also, I don't see a reason to include this change in Tests (_domain routing is a standard feature of Router, so tested there_).

Routes are protected via Authorization, but just for a project with more subdomains running one Laravel instance (_serving different content on different subdomains_), I simply like everything clean.


themsaid:
I don't see a point in adding this. sorry :)

PR #291, opened 2018-10-31
fixed request watcher overflow for large response payloads

paras-malhotra:
This PR fixes #166 and #228. Large response payloads currently throw an SQL exception due to the size exceeding the `content` column's max length and due to this exception, entries for all watchers are missed. 

This PR purges response payloads beyond a size limit that can be specified as an option to the `RequestWatcher`. This is a non-breaking change.

Issue #228, opened 2018-10-29
Large response payloads not monitored


paras-malhotra:
Any request payload that exceeds the max length of the `content` field in the `telescope_entries` table is not monitored at all.

#166 highlights this issue and I think the discussed solution is to change it mediumtext. The current size restriction for mySQL is 64KB and I think there is a strong use case to not log the response beyond that (to prevent bloating of the telescope database/table).

What's more is that if the response payload exceeds the column size, the other watchers aren't able to log either (e.g. the queries aren't logged even though they can perfectly fit the size restriction).

I think this would is would be quite a common case for users of Telescope and IMO, the Telescope storage class should handle this and possibly store a truncated snippet of the response in such cases. Also, if an insertion of an entry fails, the other entries should not be affected.

themsaid:
until we address this, feel free to run a migration to change the column type in your app.

themsaid:
Long responses will be purged now. check https://github.com/laravel/telescope/pull/291 and https://github.com/laravel/telescope/commit/a643fbd505b2359d7cf5a9be25c26589d3e9543a

PR #433, opened 2018-11-23
Changed content to longText

elimentz:
Fix for error when inserting data (Data too long)

screen shot 2018-11-23 at 11 49 54


paras-malhotra:
This was already fixed in #291. See #228. If you still want to do this in your app, you are free to change your app's migrations.

PR #436, opened 2018-11-25
Check TELESCOPE_ENABLED instead of `runningUnitTests()`

owenconti:
This is an attempt at Taylor's comment from here: https://github.com/laravel/telescope/pull/415#issuecomment-440304272

This will allow users to enable and use Telescope for unit tests.

**Telescope Changes** 

* Replace `$app->runningUnitTests()` with `config('telescope.enabled')`
* Update the migrations to check if the tables exist before attempting to create them. The reason for this change is because unit tests will often drop/re-migrate the database. However, if you configure Telescope to use a different database than your test database, the Telescope DB will not be dropped, but it will be re-migrated. 

**Application changes**

* Add new database connection for Telescope: 

```
        'telescope' => [
            'driver' => 'sqlite',
            'database' => database_path('telescope.sqlite'),
            'prefix' => '',
        ],
```

* Change my `config/telescope.php` file to use a new environment variable for the database connection: `'connection' => env('TELESCOPE_DB_CONNECTION', 'mysql'),`
* Add the new environment variable to my `.env` file: `TELESCOPE_DB_CONNECTION=telescope`
* Add new sqlite database file: `touch database/telescope.sqlite`
* Change the `Telescope::filter` closure. Calling `$this->app->isLocal()` inside the closure throws an exception which I couldn't figure out:

Change I made:

```
        $telescopeFilter = $this->app->isLocal() || $this->app->runningUnitTests();

        Telescope::filter(function (IncomingEntry $entry) use ($telescopeFilter) {
            if ($telescopeFilter) {
                return true;
            }
            
            return $entry->isReportableException() ||
                   $entry->isFailedJob() ||
                   $entry->isScheduledTask() ||
                   $entry->hasMonitoredTag();
        });
```

Exception when using `$this->app->isLocal()` inside the closure:
```
ReflectionException: Class env does not exist

/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/Container.php:779
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/Container.php:658
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/Container.php:609
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Application.php:735
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/Container.php:1222
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Application.php:495
/var/www/remoteauth-app/app/Providers/TelescopeServiceProvider.php:26
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Support/HigherOrderCollectionProxy.php:60
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Support/HigherOrderCollectionProxy.php:60
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Support/Collection.php:455
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Support/HigherOrderCollectionProxy.php:61
/var/www/telescope/src/Telescope.php:232
/var/www/telescope/src/Telescope.php:201
/var/www/telescope/src/Telescope.php:235
/var/www/telescope/src/Telescope.php:358
/var/www/telescope/src/Watchers/QueryWatcher.php:43
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:360
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:209
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:826
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:682
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:635
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:333
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Connection.php:304
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Schema/Builder.php:73
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Migrations/DatabaseMigrationRepository.php:169
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Migrations/Migrator.php:556
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Console/Migrations/MigrateCommand.php:91
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Database/Console/Migrations/MigrateCommand.php:63
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php:29
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php:87
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/BoundMethod.php:31
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Container/Container.php:572
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Console/Command.php:183
/var/www/remoteauth-app/vendor/symfony/console/Command/Command.php:255
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Console/Command.php:170
/var/www/remoteauth-app/vendor/symfony/console/Application.php:886
/var/www/remoteauth-app/vendor/symfony/console/Application.php:262
/var/www/remoteauth-app/vendor/symfony/console/Application.php:145
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Console/Application.php:89
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Console/Application.php:188
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Console/Kernel.php:250
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Testing/PendingCommand.php:136
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Testing/PendingCommand.php:218
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Testing/Concerns/InteractsWithConsole.php:55
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Testing/RefreshDatabase.php:40
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Testing/RefreshDatabase.php:17
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestCase.php:104
/var/www/remoteauth-app/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestCase.php:71
/var/www/remoteauth-app/tests/TestCase.php:15
```

**Unresolved issues/questions**

* [ ] This will enable Telescope during tests by default because Telescope is enabled by default here:  `'enabled' => env('TELESCOPE_ENABLED', true),`
* [ ] If anyone knows why the `ReflectionException: Class env does not exist` exception is thrown when calling `$this->app->isLocal()` inside the Telescope filter, that would be nice to clean up too.

### DIFF
--- a/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
+++ b/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
@@ -42,7 +42,7 @@ class CreateTelescopeEntriesTable extends Migration
                 $table->string('type', 20);
                 $table->text('content');
                 $table->dateTime('created_at')->nullable();
-    
+
                 $table->unique('uuid');
                 $table->index('batch_id');
                 $table->index(['type', 'should_display_on_index']);
@@ -53,10 +53,10 @@ class CreateTelescopeEntriesTable extends Migration
             $this->schema->create('telescope_entries_tags', function (Blueprint $table) {
                 $table->uuid('entry_uuid');
                 $table->string('tag');
-    
+
                 $table->index(['entry_uuid', 'tag']);
                 $table->index('tag');
-    
+
                 $table->foreign('entry_uuid')
                       ->references('uuid')
                       ->on('telescope_entries')
--- a/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
+++ b/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
@@ -32,37 +32,43 @@ class CreateTelescopeEntriesTable extends Migration
      */
     public function up()
     {
-        $this->schema->create('telescope_entries', function (Blueprint $table) {
-            $table->bigIncrements('sequence');
-            $table->uuid('uuid');
-            $table->uuid('batch_id');
-            $table->string('family_hash')->nullable()->index();
-            $table->boolean('should_display_on_index')->default(true);
-            $table->string('type', 20);
-            $table->text('content');
-            $table->dateTime('created_at')->nullable();
+        if (!$this->schema->hasTable('telescope_entries')) {
+            $this->schema->create('telescope_entries', function (Blueprint $table) {
+                $table->bigIncrements('sequence');
+                $table->uuid('uuid');
+                $table->uuid('batch_id');
+                $table->string('family_hash')->nullable()->index();
+                $table->boolean('should_display_on_index')->default(true);
+                $table->string('type', 20);
+                $table->text('content');
+                $table->dateTime('created_at')->nullable();
+    
+                $table->unique('uuid');
+                $table->index('batch_id');
+                $table->index(['type', 'should_display_on_index']);
+            });
+        }
 
-            $table->unique('uuid');
-            $table->index('batch_id');
-            $table->index(['type', 'should_display_on_index']);
-        });
+        if (!$this->schema->hasTable('telescope_entries_tags')) {
+            $this->schema->create('telescope_entries_tags', function (Blueprint $table) {
+                $table->uuid('entry_uuid');
+                $table->string('tag');
+    
+                $table->index(['entry_uuid', 'tag']);
+                $table->index('tag');
+    
+                $table->foreign('entry_uuid')
+                      ->references('uuid')
+                      ->on('telescope_entries')
+                      ->onDelete('cascade');
+            });
+        }
 
-        $this->schema->create('telescope_entries_tags', function (Blueprint $table) {
-            $table->uuid('entry_uuid');
-            $table->string('tag');
-
-            $table->index(['entry_uuid', 'tag']);
-            $table->index('tag');
-
-            $table->foreign('entry_uuid')
-                  ->references('uuid')
-                  ->on('telescope_entries')
-                  ->onDelete('cascade');
-        });
-
-        $this->schema->create('telescope_monitoring', function (Blueprint $table) {
-            $table->string('tag');
-        });
+        if (!$this->schema->hasTable('telescope_monitoring')) {
+            $this->schema->create('telescope_monitoring', function (Blueprint $table) {
+                $table->string('tag');
+            });
+        }
     }
 
     /**
--- a/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
+++ b/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
@@ -32,7 +32,7 @@ class CreateTelescopeEntriesTable extends Migration
      */
     public function up()
     {
-        if (!$this->schema->hasTable('telescope_entries')) {
+        if (! $this->schema->hasTable('telescope_entries')) {
             $this->schema->create('telescope_entries', function (Blueprint $table) {
                 $table->bigIncrements('sequence');
                 $table->uuid('uuid');
@@ -49,7 +49,7 @@ class CreateTelescopeEntriesTable extends Migration
             });
         }
 
-        if (!$this->schema->hasTable('telescope_entries_tags')) {
+        if (! $this->schema->hasTable('telescope_entries_tags')) {
             $this->schema->create('telescope_entries_tags', function (Blueprint $table) {
                 $table->uuid('entry_uuid');
                 $table->string('tag');
@@ -64,7 +64,7 @@ class CreateTelescopeEntriesTable extends Migration
             });
         }
 
-        if (!$this->schema->hasTable('telescope_monitoring')) {
+        if (! $this->schema->hasTable('telescope_monitoring')) {
             $this->schema->create('telescope_monitoring', function (Blueprint $table) {
                 $table->string('tag');
             });
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -103,7 +103,7 @@ class Telescope
      */
     public static function start($app)
     {
-        if ($app->runningUnitTests()) {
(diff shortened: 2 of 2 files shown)


paras-malhotra:
Firstly, I'm not sure what the migration files have to do with the goal at hand. Secondly, this change is incomplete. Simply removing `runningUnitTests` would have been fine since `registerWatchers` does not register if telescope isn't enabled. But this change makes the `registerWatchers` check useless.

Also, this is a breaking change, so should possibly target master.

owenconti:
@paras-malhotra Migration file changes are described in the description:

> Update the migrations to check if the tables exist before attempting to create them. The reason for this change is because unit tests will often drop/re-migrate the database. However, if you configure Telescope to use a different database than your test database, the Telescope DB will not be dropped, but it will be re-migrated.

I agree with the semi-redundant check in `registerWatchers`. If Telescope is disabled, should any of the static methods in `start()` be called?

I will update the target branch.

specialtactics:
Love it !

themsaid:
No need to change the migrations, I've applied the change with the config value.

owenconti:
@themsaid without the migration changes, users will get errors when using Telescope for tests if they use a separate DB for Telescope (which they should, otherwise the Telescope entries will get lost with `RefreshDatabase`)

Do you have another way around that problem?

deleugpn:
Would be nice if Telescope didn't load migrations by default and we had to run the migrations ourselves as it's a bit annoying when we're using a dedicated database and we don't want to run all migrations mixed from the project into the telescope database.

themsaid:
@owenconti honestly I don't understand why would you run Telescope in unit tests, maybe you want to debug a certain test but not the entire suit. If you are debugging a certain test just comment the RefreshDatabase and bring it back when you're done.

themsaid:
User with ID 1 can be any users created in any of the tests in your suite, that's why I think it's not very useful to run telescope for the entire test suite. I don't think you can find workable information from this.

paras-malhotra:
@owenconti you have at least 2 options if you want to do this:
1. You can use a separate telescope DB for test debugging and add the appropriate environment configs in your phpunit file
2. Instead of using the RefreshDatabase trait, you can call a custom function that doesn't drop the telescope database

owenconti:
@themsaid I think there's a misunderstanding somewhere. I'm not sure what the user ID has to do with anything. 

I think it's a valid use case to inspect and debug your application, regardless of if the requests are made during tests or not. I used it all day yesterday on my forked version without any issue, and it helps make debugging *much* easier.

@paras-malhotra your first suggestion does not work because when Laravel runs the RefreshDatabase migration, the tables already exist in the Telescope database.

Your second suggestion is valid, but why make developers jump through those hoops when an easier approach for everyone is to add a safety check in your migration?

paras-malhotra:
@owenconti Laravel does not drop all your databases when running the RefreshDatabase, it only drops the databases that are configured in your app.

https://github.com/laravel/telescope/blob/cceaf1174e1db03dc387c75622c6cdb75b54a4de/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php#L21-L26

If you see the migration above, it switches to the database connection specified in `config('telescope.storage.database.connection')`. If you change this config using environment variables defined in your `phpunit.xml` file, Laravel will only drop that database (the test Telescope database) and your original Telescope database will stay intact.

Hope this makes sense now.

owenconti:
Hi @themsaid, any luck with tracking migrations per DB connection?

themsaid:
If you use `--path` and `--database` with the migrate command you can migrate isolate Telescope migrations in a specific connection.

GuidoHendriks:
> Question is, how many developers really want to run Telescope in their test environment. I honestly don't find it very useful. - @themsaid 

What was the answer to this? Very interesting to see it's not only made possible but also enabled by default.

themsaid:
The way tests run will make it very hard for Telescope to operate, I'll need to dedicate some time to explore this but I don't think it's simple :)

GuidoHendriks:
But since this change it's enabled by default for tests:
https://github.com/laravel/telescope/commit/88d760a3931de13c10322eaeb69340f938b896c7

Doesn't seem like a good default behavior. It caused my test suite to fail because it ran out of memory. Was pretty hard to find out what the issue was.

themsaid:
@GuidoHendriks some people use it while running a single test or a test class, problems start to appear when it's running while a large test suite is being run.

GuidoHendriks:
@themsaid Doesn't feel like a good reason to change the default behavior in a patch, as it's a breaking change for people with large test suites. Allowing people to enable it for tests with a separate config value would make more sense.

telkins:
FWIW, I'm one of the developers who would very much like to use something like telescope for troubleshooting tests.  Most of my work is heavily driven by tests and having the kind of insight that telescope provides would be quite helpful.

PR #442, opened 2018-11-27
Travis already does self-update

szepeviktor:
Composer does not need self-update.
Plus some quote-normalization!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,11 +28,10 @@ services:
 
 before_install:
   - phpenv config-rm xdebug.ini || true
-  - printf "\n" | pecl install -f redis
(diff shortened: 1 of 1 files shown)


paras-malhotra:
Any reason why the laravel framework has the same composer self-update?

szepeviktor:
> Any reason why the laravel framework repo's travis file has the same composer self-update?

They may not spot the same command in Travis' log https://travis-ci.org/laravel/telescope/jobs/460436317#L480

szepeviktor:
In the near past it was a problem with Travis.

taylorotwell:
If nothing is broken I see no reason to change things.

Issue #441, opened 2018-11-27
[Feature Request] When watchers are disabled in config, remove them from the sidebar in the web UI


ryanmortier:
Hi, I ran into this today. I wanted to disable some watchers and did so with the `env()` variable but noticed the sidebar still shows the link/route. I'd open a PR but I don't know enough about Vue to do so.

Sparclex:
I think can be closed:
> No plans to add this.
https://github.com/laravel/telescope/pull/446#issuecomment-442470883

ryanmortier:
So be it.

PR #446, opened 2018-11-28
Hide disabled watchers from navigation

Sparclex:
Hides watchers from navigation if there are disabled in the config file as requested in #441.

On Problem exists, when disabling the ExceptionWatcher, the spacing between the two navigation groups disapears. I am open for suggestions that are not as ugly as the ones I am thinking of 😄

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -57,6 +57,7 @@
         <div class="row mt-4">
             <div class="col-2 sidebar">
                 <ul class="nav flex-column">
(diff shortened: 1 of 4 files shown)


taylorotwell:
No plans to add this.

mvehar:
Maybe just disabling the items (no clickable, grayed out, and tooltip that this watcher is didabled) ?

J2TeamNNL:
I think this is a feature, to know that have a lot of watcher (for whom not install this)
And you don't have to remember it

PR #452, opened 2018-11-29
[1.0] Improve performance: dont register watchers when not recording

### LINKED ISSUE #443: Improve memory utilization
It seems that when telescope is logging things that memory usage is straight up out of control.
I have a process that used to complete at my normal configuration of 512MB for the php process, and now requires over 1G to complete and it seems to be failing in memory usage.

Based on the progress bar with memory usage:
Without Telescope enabled: ~34MB
With Telescope enabled: ~414MB(run with php -dmemory_limit=1G artisan ) but dies after the command at 1G

The only unique thing about this command is that dispatches queue jobs from the command and with telescope enabled, it also takes quite some time for the command to actually complete as it just hangs after it is done. Additionally, I have this command configured as an ignored command yet it still seemingly is getting processed.

Sorry for the edit churn, I thought I had the wrong results.

paras-malhotra:
Currently, the watchers are registered and listen to events (and store entries in memory) even on ignored commands or paths. This PR improves Telescope performance by only registering watchers & event listeners on approved commands/paths. 

Resubmit of #368. The earlier PR was reverted because of a bug in failing to record failed/processed jobs. That has been fixed in this PR by adding the ability to boot watchers.

Watchers will be booted even on ignored commands and paths, but they would only be registered on approved commands and paths. 

Fixes #443

### DIFF
--- a/src/Contracts/BootableWatcher.php
+++ b/src/Contracts/BootableWatcher.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Laravel\Telescope\Contracts;
+
+interface BootableWatcher
+{
+    /**
+     * Boot the watcher.
+     *
+     * @param  \Illuminate\Contracts\Foundation\Application  $app
+     * @return void
+     */
+    public function boot($app);
+}
--- a/src/RegistersWatchers.php
+++ b/src/RegistersWatchers.php
@@ -26,9 +26,10 @@ trait RegistersWatchers
      * Register the configured Telescope watchers.
      *
      * @param  \Illuminate\Foundation\Application  $app
+     * @param  bool  $bootOnly
      * @return void
      */
-    protected static function registerWatchers($app)
+    protected static function registerWatchers($app, $bootOnly = false)
     {
         if (! config('telescope.enabled')) {
             return;
@@ -49,7 +50,13 @@ trait RegistersWatchers
 
             static::$watchers[] = get_class($watcher);
 
-            $watcher->register($app);
+            if (method_exists($watcher, 'boot')) {
+                $watcher->boot($app);
+            }
+
+            if (! $bootOnly) {
+                $watcher->register($app);
+            }
         }
     }
 }
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -107,13 +107,14 @@ class Telescope
             return;
         }
 
-        static::registerWatchers($app);
+        $needsRecording = static::runningApprovedArtisanCommand($app) ||
+            static::handlingApprovedRequest($app);
+        
+        static::registerWatchers($app, ! $needsRecording);
 
         static::registerMailableTagExtractor();
 
-        if (static::runningApprovedArtisanCommand($app) ||
-            static::handlingApprovedRequest($app)
-        ) {
+        if ($needsRecording) {
             static::startRecording();
         }
     }
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -109,7 +109,7 @@ class Telescope
 
         $needsRecording = static::runningApprovedArtisanCommand($app) ||
             static::handlingApprovedRequest($app);
-        
+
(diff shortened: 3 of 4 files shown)


deleugpn:
Quite some breaking changes here. Hard call.

paras-malhotra:
@deleugpn yes, there are some breaking changes here. Perhaps we could have this as part of a point release like v1.1.0? I couldn't rebase because the 1.0 branch is ahead of master.

deleugpn:
Laravel official packages follow semver, which means a breaking change would require Telescope 2.0.

paras-malhotra:
Hmm, the framework does have breaking changes on point releases (e.g. 5.7 to 5.8). Can't Telescope follow the same convention (1.0 to 1.1)?

taylorotwell:
What are the breaking changes? Can someone please describe them?

paras-malhotra:
@taylorotwell breaking changes is as follows:

On ignored paths and commands, earlier the `register` method of watchers was always called. Now it isn't called, instead `boot` is called if it exists on the watcher. 

For the existing watchers in the package, the final output will not change because ultimately anyway the entries were not recorded. But this may break custom watchers implemented by folks who rely on the watchers being always registered, regardless of whether it's an ignored command/path.

A typical example is JobWatcher that needs to register events for `JobProcessed` and `JobFailed` even on ignored commands (as `queue:work` is an ignored command but these events are fired in the queue work command). That's why boot is the ideal method to place such calls (without hogging memory by registering all watchers), which this PR takes care of.

I'm not sure if there is a real world example of a custom watcher that has a similar use case as JobWatcher, so maybe it's an edge case. But if one does exist, this PR would break that.

deleugpn:
On a second look, the method with signature change is `static` and it's a hard argument for extending it. The new interface provides a default implementation.
I guess it's good enough to go.

themsaid:
I don't really see any big performance wins here, what makes you think it fixes https://github.com/laravel/telescope/issues/443? 

Your PR basically doesn't register any watcher except the JobWatcher, however we need the other watchers to record related entries so it's displayed in the job preview screen.

paras-malhotra:
@themsaid imagine a long running cron job (Artisan command with task scheduling) that does thousands of queries/cache hits/mails/notifications (it's a pretty common thing to do in cron jobs I think).

All of that query/cache/mail/notification data is stored in `Telescope::$entriesQueue` only to be discarded later on. That would hog a lot of memory. Why listen to events, store it in an in-memory array and discard later? That's a huge performance loss in my books. And there's no way to prevent this because registration / listening to events / storing in-memory also happens for ignored commands.

This PR doesn't take care of the issue of recording related entries which should be an easy fix.

I'll resend a PR with this fix but if you want Telescope to be used in production, this PR is a must IMO.

themsaid:
If you put that command in the ignored list Telescope won't record anything while the command is running, thus it's not using any memory.

I don't think this PR is a must or anything, seen Spark running nicely in production without a magnificent overhead. Just use it wisely.

themsaid:
I have an idea though that might enhance the performance in another way.

paras-malhotra:
@themsaid not recording doesn't mean it's not using any memory. Even though it doesn't record, it does store all those entries in an array, which consumes memory.

themsaid:
It doesn't store anything in `Telescope::$entriesQueue` if it's not recording.

paras-malhotra:
Oh dang, I see what you mean. Hmm, I have this memory issue on a cron job on one of my production apps while running Telescope. I was assuming this is what caused it. Will re-investigate. Sorry for wasting your time 😮

themsaid:
@paras-malhotra It's ok :) Did you test your changes with the cron job? Did it solve the issue?

themsaid:
I believe this PR can help https://github.com/laravel/telescope/pull/457

PR #458, opened 2018-11-30
Add verbose times toggle

timacdonald:
This PR introduces the ability to toggle "verbose times" on index pages, allowing for more fine grained monitoring of requests.

Currently you might get 60 requests in a second OR 60 requests every second of a minute and it will appear the same on the index views. This hopes to make it easier to differentiate which - only when needed.

**TODO**
- [ ] Needs an icon

---

Here is what I've introduced...

Off (default / existing behaviour):

screen shot 2018-11-30 at 6 08 41 pm

On:

screen shot 2018-11-30 at 6 06 43 pm

## Reasoning

A Laravel app I'm working with is serving an API that is used by an iOS and an Android app. The mobile app makes a *heap* of requests at once and I'm currently using Telescope to monitor those requests in hopes I can get the mobile dev to chill out with all the dang requests (at one point it was over 25 a second every time the user navigated 😳) but also to stupid check everything else like cache hits etc.

I was struggling to keep track of the large number of requests being made as the "Happened" time only shows minute increments. I needed more precise time information on the index page to see when those requests are coming through. I know that verbose time is available in the preview screen, but I needed it specifically on the index page as I am watching the requests in real-time.

## Technical notes

- I didn't disable the `updateTimeAgo`. It still runs in the background, but I figure as it won't find any elements with the `data-timeago` attribute it isn't going to matter performance wise. Happy to disable / enable though if you wanted.
- I made the  `penultimate-column-right` class (great naming there!) only apply when the terse times are shown so that the label is left aligned with the longer times.

Hopefully this seems useful - it really helped me to understand how the requests were coming in.

---
Telescope is a fantastic tool, thank you for making and maintaining it!

### DIFF
--- a/resources/js/mixins/indexScreens.js
+++ b/resources/js/mixins/indexScreens.js
@@ -1,0 +1,40 @@
+import moment from 'moment-timezone';
+
+export default {
+    data() {
+        return {
+            displayVerboseTimes: localStorage.displayVerboseTimes === '1',
+        }
+    },
+
+    methods: {
+        /**
+         * Toggle displaying verbose or terse times.
+         */
+        toggleVerboseTimes(){
+            if (!this.displayVerboseTimes) {
+                this.displayVerboseTimes = true;
+                localStorage.displayVerboseTimes = 1;
+            } else {
+                this.displayVerboseTimes = false;
+                localStorage.displayVerboseTimes = 0;
+            }
+        },
+
+        /**
+         * Format to verbose time
+         */
+        verboseTime(time){
+            return moment(time + ' Z').utc().local().format('DD/MM/YY HH:mm:ss');
+        },
+    },
+
+    computed: {
+        /**
+         * Check if terse times should be displayed.
+         */
+        displayTerseTimes(){
+            return ! this.displayVerboseTimes;
+        },
+    },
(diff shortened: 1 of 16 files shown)


themsaid:
I don't think showing the seconds would make any difference, the requests are ordered based on time so you know which request came first and you know how longs passed since each request, do you know the timestamp of each request before hand? So you know you want to look at the request that came at this exact second?

taylorotwell:
I don't think I would put the button up there in the top row. I would put it in the table header of the Happened column.

ragingdave:
Any reason not to just make it a tooltip instead? Rather than clutter up the UI with more buttons/toggles?

themsaid:
screen shot 2018-12-01 at 9 57 06 am


I've pushed an update that shows the number of seconds elapsed instead of "Just Now".

timacdonald:
Hey @themsaid thanks very much for your work on this. I totally understand not wanting to add configuration and wanting to keep it simple. The last thing I want to do is complicate it. Simplicity is key!

I really appreciate your compromise on this - I will find this change *very* handy and I think your solution at the end of the day is better than my approach for a wider use-case.

Thank you!

Issue #444, opened 2018-11-28
502 Bad Gateway when enabled


austenc:
Hello, 

When telescope is enabled, one of the routes in our app gives an nginx 502 bad gateway error. We experience this on a forge server _and locally_ on valet. Without telescope, this route works fine. It is fairly resource intensive, but it does work. Tried (unsuccessfully) to debug this in various ways but haven't come up with much. Strangely, commenting out everything in the `register()` method of my app's `TelescopeServiceProvider` doesn't change anything. Toggling the `TELESCOPE_ENABLED` env variable does fix the issue though. 

Nothing goes in the `laravel.log` file, nothing in the nginx error log, and the php-fpm log only shows:
![image](https://user-images.githubusercontent.com/575421/49119711-a5f9d600-f266-11e8-9140-4f98c313fc8e.png)

What else might I try to debug this?

austenc:
I've narrowed this down a bit more. It seems it has something to do with this line in particular:
https://github.com/laravel/telescope/blob/1.0/src/Watchers/RequestWatcher.php#L179 

The data I have is a collection of 6 eloquent models, each with 7 relations loaded _some of which also load additional relations themselves_. I suspect something in this nesting or total string/data length is causing the issue with the json encoding. Will report back as I discover more.

themsaid:
How did you reach to the conclusion that the issue is in the request watcher?

austenc:
@themsaid Due to the cryptic nature of this error (no nginx/php logs to work from) I did it the old fashioned way -- commenting out various things in telescope code until I started seeing my route work without a 502. This led me to the request watcher and that line in particular. 

If I modify that line to something like this, it begins working again:
```
'properties' => json_decode(json_encode($value->take(3)), true),
```
This is **not** a solution, I just did it as an example. If I increase the number to 5+, it causes nginx to return a 502 again. Perhaps the array/object is just too big to serialize in this case? 

Happy to show you on a google Hangout if you'd like to pair up to investigate. Is trimming / limiting the data that gets sent to `json_encode` in certain cases something that you'd consider adding? Can to give it a go and submit a PR if so. 

Oh... and Telescope is awesome by the way! Thanks for sharing such a great package with the community!

austenc:
I believe this comes down to the view having too many nested relations that are being serialized. Even calling `toArray()` on the `$value` is causing the 502. Thoughts on how we might limit the length / size of view data telescope records? 

[...]

PR #460, opened 2018-11-30
Purge view data

perifer:
Related to https://github.com/laravel/telescope/pull/291 - I have the same problem but for views with a lot of data, for example a bunch of models.

Not super happy for the solution - `json_encode()` on a large body of data is probably not for free and feels a bit odd to do the purging in multiple places.

### DIFF
--- a/src/Watchers/RequestWatcher.php
+++ b/src/Watchers/RequestWatcher.php
@@ -140,9 +140,11 @@ class RequestWatcher extends Watcher
         }
 
         if ($response instanceof IlluminateResponse && $response->getOriginalContent() instanceof View) {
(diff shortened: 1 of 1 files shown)


themsaid:
See https://github.com/laravel/telescope/issues/444#issuecomment-443225529

We might just show the class name but not try to display properties or content as json_encode will cause a memory issue.

perifer:
Just a side-note: the current purge implementation that this was inspired of probably have similar memory issued - although easier to fix. See https://github.com/laravel/telescope/blob/1.0/src/Watchers/RequestWatcher.php#L136 where `json_decode()` is done before `contentWithinLimits()`. It's also done twice.

PR #462, opened 2018-12-01
[1.0] Provide AfterRecording Hook

deleugpn:
This pull request adds a `afterRecording` hook on Telescope. This will allow the developer to do whatever he/she wants when Telescope makes a new Entry. There could be lots of possibilities since the callback allows the developer to implement any logic desired.
The driver for this pull request for me is a long-running background process that does too many things.
I would still like to be able to record everything that the background job is doing for later analyzes. However, the background process can run for up to 30 minutes and can perform millions of actions.
With an `afterRecording` hook, I can monitor Telescope and force it to store and flush after each 20'000 records. This will free up memory and avoid building up more entries than the Insert Query can handle.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -35,6 +35,13 @@ class Telescope
     public static $filterBatchUsing = [];
 
     /**
+     * The callback executed after queuing a new entry.
+     *
+     * @var \Closure
+     */
+    public static $recordingCallback;
+
+    /**
      * The callback that adds tags to the record.
      *
      * @var \Closure
@@ -241,6 +248,10 @@ class Telescope
         static::withoutRecording(function () use ($entry) {
             if (collect(static::$filterUsing)->every->__invoke($entry)) {
                 static::$entriesQueue[] = $entry;
+            }
+
+            if (static::$recordingCallback) {
+                call_user_func(static::$recordingCallback, new static);
             }
         });
     }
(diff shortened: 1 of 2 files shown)


themsaid:
@deleugpn I think this is useful yes but in your case the storage will be on several batches so for example if it's a console command, the entries will be stored with the console command entry, so if you check the entry you can find all related entries under that. For the stores that you'll do later, they won't be grouped under that console command, so it'll be very hard to keep track of what's going on.

deleugpn:
I did consider that fact and I still prefer it in different batches than not working at all. Right now my production system is just failing to store the entries entirely.
There's one particular case that it was even causing memory exhaustion and killing the process.

deleugpn:
I also think it's worth to send a separate PR to promote the batch uuid into a static variable defined on start-up so it only changes if the Framework shuts down.

src/Telescope.php
```diff
@@ -242,6 +249,10 @@ protected static function record(string $type, IncomingEntry $entry)
             if (collect(static::$filterUsing)->every->__invoke($entry)) {
                 static::$entriesQueue[] = $entry;
             }
+
+            if (static::$afterRecordingHook) {
+                call_user_func(static::$afterRecordingHook, new static);
```
MatanYadaev, line 254: Why did you pass a `new static` parameter instead of the `$entry`?

PR #466, opened 2018-12-02
[1.0] Generate batch uuid at start up

deleugpn:
As discussed on https://github.com/laravel/telescope/pull/462 with @themsaid, it might be an easy win to only generate the `batch uuid` at Telescope Start up and keep it static throughout the entire lifecycle. If the framework gets terminated and started again, we then get a new uuid.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -21,6 +21,13 @@ class Telescope
         RegistersWatchers;
 
     /**
+     * The uuid for the process to be used when storing entries.
+     *
+     * @var string
+     */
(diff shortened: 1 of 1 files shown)


themsaid:
@deleugpn I wonder what's your use case for this?

deleugpn:
This is mostly to address your concerns on #462. It's just an easy commodity to be able to save all entries with the same uuid. But feel free to close this one. As I said on #462, I'd rather have it working even if I don't get the same batch id for each entry than not working at all.

PR #482, opened 2018-12-08
Fixing display wrong timestamp

### LINKED ISSUE #429: Displayed timestamp not honoring timezone
I see that there are a number of different issues reported on the same subject, but I dont think it has been "fixed" per se

As you can see based on the attached screenshots, the timestamp in the log file is not matching what is displayed (the value in the log file is correct, it is using the value set in `config/app`

Every Telescope entry is displayed with the wrong offset, not just Logs (that is just the easiest way to show what issue I am seeing)

image

phannaly:
Timezone works perfectly when retrieving but it shows the wrong timezone when Vue render template using `localTime` function. It will fix #429

### DIFF
--- a/resources/js/base.js
+++ b/resources/js/base.js
@@ -50,7 +50,7 @@ export default {
          * Show the time in local time.
          */
         localTime(time){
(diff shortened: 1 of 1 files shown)


phannaly:
@themsaid what's wrong with this?

PR #489, opened 2018-12-10
Minimum Log Level for LogWatcher

Jason-Morcos:
It would be useful to be able to set a log level config option for the LogWatcher to make the LogWatcher only log messages greater than a certain log level. In many cases, one may want to have a separate log level for Telescope than for their core application. This PR creates a "level" config option for LogWatcher and uses core Laravel's ParsesLogConfiguration trait to compare log levels. Only log messages with a log level greater than or equal to the passed in level would be logged in Telescope.

There probably needs to be some testing written around this, but I'm having a little trouble figuring out the best way to test that an entry is _not_ written into the database.


themsaid:
You can filter the entry by the "level" parameter in your `TelescopeServiceProvider` and only log what you need.

PR #524, opened 2018-12-31
Check if tables already exists

diogogomeswww:
The Telescope migrations are always running for the same connection: 
`config('telescope.storage.database.connection')`

So when we run a migration for a different connection we get an error saying:
`Base table or view already exists: 1050 Table 'telescope_entries' already exists`

On this PR I've added a check if the table already exists that will prevent this error.


themsaid:
This is not correct, please check the forums or previous issues for help and share the command you use to run the migrations on a different connection.

Issue #504, opened 2018-12-13
Mailable ExtractTags BadMethodCallException


Jason-Morcos:
A job attempts to send an email that extends Illuminate's Mailable and uses Queueable, and SerializesModels. The following crash occurs.

Telescope v1.0.9

```

BadMethodCallException: Method Illuminate\Database\Eloquent\Collection::getKey does not exist. in /vendor/laravel/framework/src/Illuminate/Support/Traits/Macroable.php:100

Stack trace:
#0 /vendor/laravel/telescope/src/FormatModel.php(18): Illuminate\Support\Collection->__call('getKey', Array)
#1 /vendor/laravel/telescope/src/ExtractTags.php(30): Laravel\Telescope\FormatModel::given(Object(Illuminate\Database\Eloquent\Collection))
#2 [internal function]: Laravel\Telescope\ExtractTags::Laravel\Telescope\{closure}(Object(Illuminate\Database\Eloquent\Collection), 1)
#3 /vendor/laravel/framework/src/Illuminate/Support/Collection.php(1007): array_map(Object(Closure), Array, Array)
#4 /vendor/laravel/telescope/src/ExtractTags.php(31): Illuminate\Support\Collection->map(Object(Closure))
#5 /vendor/laravel/telescope/src/ExtractsMailableTags.php(19): Laravel\Telescope\ExtractTags::from(Object(App\Mail\CustomMailable))
#6 [internal function]: Laravel\Telescope\Telescope::Laravel\Telescope\{closure}(Object(App\Mail\ CustomMailable))
#7 /vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(284): call_user_func(Object(Closure), Object(App\Mail\CustomMailable))
#8 /vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(153): Illuminate\Mail\Mailable->buildViewData()
#9 /vendor/laravel/framework/src/Illuminate/Support/Traits/Localizable.php(19): Illuminate\Mail\Mailable->Illuminate\Mail\{closure}()
#10 /vendor/laravel/framework/src/Illuminate/Mail/Mailable.php(160): Illuminate\Mail\Mailable->withLocale(NULL, Object(Closure))
#11 /vendor/laravel/framework/src/Illuminate/Mail/Mailer.php(274): Illuminate\Mail\Mailable->send(Object(Illuminate\Mail\Mailer))
[...]

PR #529, opened 2019-01-10
ExtractTags from multi-level Collections

JeremyJass:
Solution to Issue #504.

The problem arises when you take a simple Collection of Models, and execute a 'groupBy' or otherwise modify the Collection so it is no longer single-level. In the ExtractTags file, modelsFor() was assuming that if a property of the target class was an EloquentCollection that it was always a simple Collection of Models.

There are 'flatten' methods available to the Collection class, but I wasn't completely sure how they would handle variable depth Collections, or if you had a mixed Collection where only some elements were instances of the Model class.

### DIFF
--- a/src/ExtractTags.php
+++ b/src/ExtractTags.php
@@ -175,6 +175,7 @@ class ExtractTags
                 $flat_array = array_merge($flat_array, [$item]);
             }
         });
+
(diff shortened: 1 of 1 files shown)


taylorotwell:

Thanks!

PR #546, opened 2019-01-30
Make longer content entries possible

mrk-j:
I encountered an issue with Telescope when logging requests with large `Cache` payloads. The requests would not show up in the dashboard. When using a longer field-type for the content the request was logged as it should.


bonzai:
#433

PR #548, opened 2019-02-01
prevent heavy memory usage by saving query-log-entries in chunks into the db

### LINKED ISSUE #443: Improve memory utilization

dasraab:
this should also fix #443

### DIFF
--- a/src/Storage/DatabaseEntriesRepository.php
+++ b/src/Storage/DatabaseEntriesRepository.php
@@ -111,11 +111,15 @@ class DatabaseEntriesRepository implements Contract, ClearableRepository, Prunab
 
         $this->storeExceptions($exceptions);
 
-        $this->table('telescope_entries')->insert($entries->map(function ($entry) {
-            $entry->content = json_encode($entry->content);
+        $table = $this->table('telescope_entries');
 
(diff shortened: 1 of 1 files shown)


drsdre:
Maybe it makes sense to make the chunk size configurable to prevent OoM on servers with less memory. Or event better, detect when memory is getting low and start saving data to DB to free up memory.

VictorProductFlow:
Regression on 5.4

Telescope still consumes a lot of memory unless I disable QueryWatcher.

Issue #540, opened 2019-01-23
Telescope ignores migrate's --database parameter


loykianos:
Whenever I try using artisan migrate or migrate:refresh with the --database parameter (f.e. different testing database), telescope produces a "duplicate table" error. 

> Migrating: 2018_08_08_100000_create_telescope_entries_table
> 
> Illuminate\Database\QueryException  : SQLSTATE[42P07]: Duplicate table: 7 ERROR:  relation "telescope_entries" already exists (SQL: create table "telescope_entries" ("sequence" bigserial primary key not null, "uuid" uuid not null, "batch_id" uuid not null, "family_hash" varchar(255) null, "should_display_on_index" boolean not null default '1', "type" varchar(20) not null, "content" text not null, "created_at" timestamp(0) without time zone null))

simioluwatomi:
I just wanted to report this. It is becoming annoying having to fiddle with Telescope's database connection value in `config/telescope` everytime I want to migrate my testing database.

Screenshot below

![image](https://user-images.githubusercontent.com/22191563/51620440-acd55a00-1f32-11e9-86f6-401e95807e14.png)

diogogomeswww:
My solution was to create an `.env.testing` with the entries:
```
APP_ENV=testing
DB_CONNECTION=mysql_testing
TELESCOPE_ENABLED=0
```

And then run:
`art migrate:fresh --database=mysql_testing --env=testing`

that `env` flag will load the `.env.testing` file and will fix the issue with telescope migrations

tim-mortimer:
@diogogomeswww nice work! I'm assuming that means you don't need to specify the `--database=mysql_testing` option when running the command then?

diogogomeswww:
@tim-mortimer ah yes, you are right, of course:
`art migrate:fresh --env=testing` is enough.
(going to edit above)

ntraykov:
I also have this issue. I also reached the point where the only solution that I have was to create an .env.testing file, but still, this shouldn't work like that, because it is confusing. Telescope should respect the --database parameter.

AmirrezaNasiri:
If you don't want to create new `.env.testing` file which may conflict with what you've already set up in `.env` or you're simply not comfort with that, you can make telescope database selection conditional. Here it is:
```php
// config/telescope.php
return [
   // ...
   'storage' => [
      'database' => [
         'connection' => 'testing' === env('APP_ENV') ? env('TEST_DB_CONNECTION') : env('DB_CONNECTION'),
[...]

PR #551, opened 2019-02-04
Don't run migrations when connection differs from configured connection

tim-mortimer:
This is my proposed fix to https://github.com/laravel/telescope/issues/540

When a user runs `php artisan migrate --database=mysql_testing` (or any other migrate command) whilst specifying the database connection to use through the --database option, Telescope ignores the specified database connection and attempts to migrate the database on the connection configured in config/telescope.php.

This PR checks the connection the Migrator is currently using and does not run Telescope's migrations if it differs from the connection configured in config/telescope.php.

Hope this helps!

### DIFF
--- a/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
+++ b/src/Storage/migrations/2018_08_08_100000_create_telescope_entries_table.php
@@ -32,6 +32,10 @@ class CreateTelescopeEntriesTable extends Migration
      */
     public function up()
     {
+        if (!app()->environment('self-testing') && !$this->currentConnectionIsConfiguredTelescopeConnection()) {
+            return;
+        }
+
         $this->schema->create('telescope_entries', function (Blueprint $table) {
             $table->bigIncrements('sequence');
             $table->uuid('uuid');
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure I totally understand `self-testing` environment?

tim-mortimer:
@taylorotwell, in the case where the Telescope test suite is running, it is the one time that Telescope uses a different connection (the testing connection) to the one configured in config/telescope.php, and thus you still want the migrations to run. However, outside of the test suite, you want Telescope to use the connection configured in config/telescope.php. Does that make sense? Apologies if I'm missing something!

simioluwatomi:
What's the status on this please?

diogogomeswww:
My solution was: https://github.com/laravel/telescope/issues/540#issuecomment-484847572

mabasic:
Why is this PR closed and not worked on?

There is a very annoying open issue #540 which is solved with this PR.

When I run `php artisan migrate --database=x` I expect that all migrations are executed on database connection `x`.

What happens is that all migrations except telescope migrations are executed on database connection `x`, and telescope migrations are executed on the database connection specified in the `.env` file.

Issue #500, opened 2018-12-13
TelescopeServiceProvider not registered in config/app.php


mavew:
Using Laravel 5.7.17.

On a fresh install, when I run `composer require laravel/telescope`, followed by `php artisan telescope:install`, I don't see the `TelescopeServiceProvider` provider registered in my app.php file. After migrating the database, the `/telescope` route works just fine, though I'm unable to figure out why it doesn't register itself in the config file, because the documentation states that it should.

OS: Windows 10
![capture](https://user-images.githubusercontent.com/12530941/49925287-8ae8c080-ff0c-11e8-87fe-ea55c59625c7.PNG)

themsaid:
Look closer. If it's working then it's registered.

mavew:
I installed it on a fresh Laravel install just now, and confirmed that it was working. Here's my `app.php` file.

`<?php

return [

    /*
    |--------------------------------------------------------------------------
    | Application Name
    |--------------------------------------------------------------------------
    |
    | This value is the name of your application. This value is used when the
    | framework needs to place the application's name in a notification or
    | any other location as required by the application or its packages.
    |
    */

    'name' => env('APP_NAME', 'Laravel'),

    /*
    |--------------------------------------------------------------------------
    | Application Environment
    |--------------------------------------------------------------------------
    |
    | This value determines the "environment" your application is currently
    | running in. This may determine how you prefer to configure various
    | services the application utilizes. Set this in your ".env" file.
    |
    */

    'env' => env('APP_ENV', 'production'),

    /*
    |--------------------------------------------------------------------------
    | Application Debug Mode
    |--------------------------------------------------------------------------
    |
    | When your application is in debug mode, detailed error messages with
    | stack traces will be shown on every error that occurs within your
    | application. If disabled, a simple generic error page is shown.
    |
    */

    'debug' => env('APP_DEBUG', false),

    /*
    |--------------------------------------------------------------------------
    | Application URL
    |--------------------------------------------------------------------------
    |
    | This URL is used by the console to properly generate URLs when using
    | the Artisan command line tool. You should set this to the root of
    | your application so that it is used when running Artisan tasks.
    |
    */

    'url' => env('APP_URL', 'http://localhost'),

    'asset_url' => env('ASSET_URL', null),

    /*
    |--------------------------------------------------------------------------
    | Application Timezone
    |--------------------------------------------------------------------------
    |
    | Here you may specify the default timezone for your application, which
[...]

Issue #271, opened 2018-10-30
Documentation improvement


cosecantt:
Mentioning 
````For non-local usage authorization, add App\Providers\TelescopeServiceProvider::class to config/app.php ````
is missing in the documentation.

Te7a-Houdini:
i think it's added when running ```php artisan telescope:install```

cosecantt:
For laravel 5.7 it did not add and did not work until I added it manually.

bgarrison25:
by "did not work" do you mean you couldn't even get to /telescope or that it wasn't recording things?

cosecantt:
By did not work I meant it did not work as expected. When logged in I
visited /telescope and it redirected to 403 unauthorized page. Once I added
above class to the config without any other changes it worked as it
expected.

bgarrison25:
Yeah, I don't know why it didn't add that provider class to config/app.php for you but in the three installs I have done so far it did it automatically. I can't reproduce it myself.

robertmylne:
Didn't add it to the file for me either. Only did when I ran on production server.

Issue #178, opened 2018-10-25
Authorization fails


eossa:
I change the gate method in TelescopeServiceProvider:

```
Gate::define('viewTelescope', function ($user) {
    return in_array($user->username, [
        '1234567890',
    ]);
});
```

Instead of use the email attribute I use the username attribute for login, however in production returns a 403 response when I am logged with the user with that username.

fridzema:
By default the laravel users table uses the `name` column for username, do you have changed this?
Have you tried to access before changing it?

themsaid:
That's something in your code, you'll need to look deeper.

eossa:
I found the trouble, I had not register the **TelescopeServiceProvider** in my *config/app.php*.

struyfv:
Hi, I have the same issue and I registered the TelescopeServiceProvider.

My code:

```
protected function gate()
    {
//        dd('test');
        Gate::define('viewTelescope', function ($user) {
            dd($user);
            return in_array($user->email, [
                'email'
            ]);
        });
    }
```
When I uncomment dd('test'); I see it, but the code never stops at dd($user);

Any idea?

baekerIT:
Hi,

i’ve a solution, but don’t have the time to look now. This evening I  will look through my current project, where the problem has caused.

struyfv:
I found out why it didn't work. I didn't use the default 'web' guard for auth.

luqmanrom:
@baekerIT I don't think you need to remove `Authorize::class`. The only time it will allow only locals if your Telescope auth callback is not defined. If your TelescopeServiceProvider is properly defined, the callback and the viewTelescope gate will be defined there. 


Make sure these things:

1. Include `TelescopeServiceProvider` in `config/telescope.php`
2. Clear the config cache - `php artisan config:clear`
3. Clear the compiled files - `php artisan clear-compiled`

Hope it works

swilla:
I was fighting with this for a while, and what I needed to do was remove `laravel\telescope` from the `dont-discover` list in composer.json, then all the conditional loading describe in the documentation worked as expected.

mateo2181:
How can I include TelescopeServiceProvider in config/telescope.php?

fridzema:
@mateo2181 you must include it in `config/app.php`

mateo2181:
Yes, I did it and works. Thanks!

MutedFaith:
> I've got following Solution for reach Telescope in Production:
> 
[...]

PR #558, opened 2019-02-12
Fix config/app.php linendings

rolandstarke:
If you checkout a new laravel project on windows your `config/app.php` might end up with linux lineendings.

If you then run `php artisan telescope:install` this string can't be found in the config: `"{$namespace}\\Providers\EventServiceProvider::class,".PHP_EOL`. And so the `TelescopeServiceProvider` is not added to it.

This caused https://github.com/laravel/telescope/issues/178, https://github.com/laravel/telescope/issues/271, https://github.com/laravel/telescope/issues/500 and probably some more issues like "darkmode not working", when the underlying problem was that the TelescopeServiceProvider wasn't registered.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -61,8 +61,8 @@ class InstallCommand extends Command
         }
 
         file_put_contents(config_path('app.php'), str_replace(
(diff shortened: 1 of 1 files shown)


taylorotwell:
This may be a better candidate for 1.0 branch if it is a bug fix?

rolandstarke:
Yes

Issue #567, opened 2019-02-27
Error when using Date::use(CarbonImmutable::class)


kbond:
When using `Date::use(CarbonImmutable::class)` I get the following error when trying to load telescope in the browser:

```
Argument 7 passed to Laravel\Telescope\EntryResult::__construct() must be an instance of Carbon\Carbon, instance of Carbon\CarbonImmutable given, called in /vendor/laravel/telescope/src/Storage/DatabaseEntriesRepository.php on line 93
```

virp:
The same problem, I think you need to change Carbon to CarbonInterface type hint in Telescope package

themsaid:
This was fixed on the 2.0 branch.

If you're using Laravel 5.8 please use Telescope 2.0

PR #570, opened 2019-03-01
Change Carbon type hint to CarbonInterface

virp:
Change Carbon type hint to CarbonInterface for CorbonImmutable compatibility.
A possible solution to the problem described in this issue #567


jpmurray:
One would need to make the same change in `2.0` also. I guess it's better to submit another PR for `2.0` branch?

PR #579, opened 2019-03-06
Add tag for http status codes

oad-dwarf:
@themsaid here you go :-)

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -434,6 +434,7 @@ class Telescope
      */
     public static function recordRequest(IncomingEntry $entry)
     {
+        $entry->tags[] = 'Status:' . $entry->content['response_status'];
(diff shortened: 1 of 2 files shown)


themsaid:
You'll need to do it in the RequestWatcher itself, check the other watchers as an example.

oad-dwarf:
Thank you, I have made the change.

taylorotwell:
I don't want to tag on status code. It's just a very general tag.

oad-dwarf:
Adding the http status code as a tag makes it easy to find requests that return 422 or 500 without scrolling a lot.

This is our use case, which I think is a very nice feature for a tool such as Telescope. Would you consider taking in this PR on that basis?

Or do you have plans for another way to search for requests on status code?

PR #591, opened 2019-03-18
Use a custom model key

kainxspirits:
It would be nice to be able to customize the model key that will be displayed on the dasbhoard without having to change the getKey method.

I don't like much the idea of adding a method on the model, but it's optional . (I thought of adding this in the config, but it looks cleaner directly on the model)


taylorotwell:

PR #597, opened 2019-03-23
Showing more information (class, method and args) in Excpetion details screen

pavankumarkatakam:
That information is very useful, so it needs to be in the UI.

### DIFF
--- a/resources/js/components/Stacktrace.vue
+++ b/resources/js/components/Stacktrace.vue
@@ -1,41 +1,75 @@
 <script type="text/ecmascript-6">
-    import _ from "lodash"
+import _ from "lodash";
 
-    export default {
-        props: ['trace'],
+export default {
+  props: ["trace"],
 
-        /**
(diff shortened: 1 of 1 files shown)


taylorotwell:

Thanks!

pavankumarkatakam:
Im suprised to see this one is closed with out merging.

Explanation: In the stacktrace we are showing only the file path and line numbers, but in some situation like Fatal errors this will not give information to correct the issue. In that cases we need which method is calling and what are the arguments are. One example recently i got stuck AWS SES exception saying the address is not not compliance with RFC thing and after logging it seems that user have given an email address that doesn't have dot.  If im able to see the argument of variable **to** going to mail::send then it is not required me to keep additional line **\Log::error()**.

Tests: I have seen the tests folder and there is no UI related tests to me to continue. But i have only changed the Stacktrace.vue and not any PHP components.

PR #604, opened 2019-03-29
Avoid error exception if currenty entry has empty content

Skullbock:
Small update to avoid strange, but possible error when for some reason the current entry content is null
`array_merge(): Argument #1 is not an array`


taylorotwell:
It can be fixed with just a single short ternary. I'll fix it.

Skullbock:
Ok, thanks!

PR #608, opened 2019-04-03
Change telescope connection default

mathieutu:
Hi, 
The default database connection of telescope is a copy of initial database.php one.

However, if a developer changes the default connection in database.php config file, telescope will be wrong. With this modification, Telescope will always use by default the default connection of the project.

Thanks.
Matt'

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -46,7 +46,7 @@ return [
 
     'storage' => [
         'database' => [
-            'connection' => env('DB_CONNECTION', 'mysql'),
+            'connection' => config('database.default'),
         ],
     ],
 


deleugpn:
I'm not sure if it's wise to use the `config` component from a config itself. I don't remember if they get filled incrementally or all at once, but I'm sure the order of process would impact the availability of the information.

taylorotwell:
Agree with @deleugpn

mathieutu:
Yeah actually I've took that in consideration, and the config files are processed alphabetically. Telescope will always been loaded after database, and it works without problem (I'm actually doing that on all my projets).

Bantes:
my different projects use the same database, when i enable telescope at two project, i want to record at different database. how can i do for this?

Issue #614, opened 2019-04-20
Too many parameter bindings during insert when using sqlsrv database driver


Namoshek:
- Telescope Version: 2.0.2
- Laravel Version: 5.8.7
- PHP Version: 7.2.13
- Database Driver & Version: sqlsrv & pdo_sqlsrv  @ v5.3.0

### Description:

When performing heavy operations during a request or command execution where too many  recordable actions are performed, it happens very frequently that Telescope is not able to store the captured entries. Recordable actions may be events, logs or any other watcher objective.

The capturing itself seems to work fine (even when capturing 10k+ actions), but when storing the recordings, nothing is stored and Telescope crashes with the error:
> PDOException: SQLSTATE[IMSSP]: Tried to bind parameter number 2101.  SQL Server supports a maximum of 2100 parameters.

### Steps To Reproduce:

- Use sqlsrv database driver (probably the only one with a hard-limit on bindable parameters).
- Fire enough events or perform enough other recordable stuff during a request or command execution.

### Initial Analysis:

My guess is that this simply happens because Telescope tries to store all recordings right before termination of the request/command:

https://github.com/laravel/telescope/blob/92bc1f7566b8f0cecef036129caf8e9451b7b117/src/ListensForStorageOpportunities.php#L41-L46

https://github.com/laravel/telescope/blob/92bc1f7566b8f0cecef036129caf8e9451b7b117/src/Telescope.php#L540-L565

https://github.com/laravel/telescope/blob/92bc1f7566b8f0cecef036129caf8e9451b7b117/src/Storage/DatabaseEntriesRepository.php#L104-L125

The rather obvious workarounds would be to:
1. store recordings as soon as the `$entriesQueue` reaches a certain limit
2. determine the ideal chunk size based on the used database driver
3. reduce the chunk size used by the `DatabaseEntriesRepository`

As 1) is a design change and most likely BC-breaking and 2) could still lead to undesired results in the future (or with custom database drivers), I'd aim for 3) and make the chunk size configurable. This way it remains BC and sqlsrv users can fix the issue themselves. A chunk size of around 250 should fix the issue as the `telescope_entries` table has 8 columns (even though not all 8 are given during an insert).

FrittenKeeZ:
The chunk size won't fix it alone, since you can end up with a single entry having over 2100 bindings.
I don't know if it's even possible to fix.

Namoshek:
Now I'm curious how a single entry can have over 2100 bindings...?

[...]

PR #615, opened 2019-04-21
Make chunk size of database storage driver configurable

Namoshek:
The PR introduces a new `chunk_size` configuration parameter for the `database` storage driver in order to fix the issue explained in #614.

In case someone created a custom version of the `TelescopeServiceProvider` which does not pass the new `$chunkSize` parameter to the database repository, the repository will still work due to the old default being used. Also users relying on the default service provider will have no issues as the old default is used if no configured chunk size is being found.

What I'm not sure about is how to properly document the setting. Personally, I'd prefer if there was a thorough explanation in `config/telescope.php`. There isn't really a common way to document nested configuration options though. Alternatively, we can also add it to the official documentation. Basically, I'm open to suggestions.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -47,6 +47,7 @@ return [
     'storage' => [
         'database' => [
             'connection' => env('DB_CONNECTION', 'mysql'),
+            'chunk_size' => 1000,
         ],
     ],
 
--- a/src/Storage/DatabaseEntriesRepository.php
+++ b/src/Storage/DatabaseEntriesRepository.php
@@ -22,6 +22,13 @@ class DatabaseEntriesRepository implements Contract, ClearableRepository, Prunab
     protected $connection;
 
     /**
+     * Number of recorded entries that will be inserted at once into the database.
+     *
+     * @var int
+     */
+    protected $chunkSize = 1000;
+
+    /**
      * The tags currently being monitored.
      *
      * @var array|null
@@ -32,11 +39,16 @@ class DatabaseEntriesRepository implements Contract, ClearableRepository, Prunab
      * Create a new database repository.
      *
      * @param  string  $connection
+     * @param  int     $chunkSize
      * @return void
      */
-    public function __construct(string $connection)
+    public function __construct(string $connection, int $chunkSize = null)
     {
         $this->connection = $connection;
+
+        if ($chunkSize) {
+            $this->chunkSize = $chunkSize;
+        }
     }
 
     /**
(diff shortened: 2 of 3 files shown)


delta1186:
**Question**: @taylorotwell when will this merge with be rolled up in a new release?  Thanks, Kevin

The rest are observations from my experiences with trying to get around this issue today and a possible bug or user error. :)

Ironically I ran into this issue today and after a few hours of messing with various telescope config options for ignoring a custom console command, which I wasn't able to get to work.  

protected $signature = '**connect:updateAdLists**';  this is from my custom console command class

I was trying to add connect:updateAdLists to either of the following areas.  Used lower case and how it is in the $signature var.    

The console command just dispatches a job so I also tried to set the following env vars, but I still got the following error as reported and fixed in this PR.  

TELESCOPE_COMMAND_WATCHER=false
TELESCOPE_JOB_WATCHER=false

SQLSTATE[IMSSP]: Tried to bind parameter number 2101.  SQL Server supports a maximum of 2100 parameters. (SQL: insert into [telescope_entries]....

```
'ignore_commands' => [
        'connect:updateAdLists'
    ],

Watchers\CommandWatcher::class => [
            'enabled' => env('TELESCOPE_COMMAND_WATCHER', true),
            'ignore' => ['connect:updateAdLists'],
        ],
```
So I wanted to mention the above in case something is not working as expected in telescope.  Hopefully I am just missing something with the ignore options, but I don't think I am especially when I am supposedly turning the watchers off by setting those env vars to false and telescope still errors because it tries record those entries.

Namoshek:
You can always pull in a specific commit through composer by specifying the branch and commit instead of a version:
```json
"require": {
    "laravel/telescope": "2.0#1749dae1"
}
```
This will pull branch `2.0` at commit `1749dae1` (most recent).

The recent changes since the last version can be found here: [v2.0.4...2.0](https://github.com/laravel/telescope/compare/v2.0.4...2.0).

FrittenKeeZ:
> You can always pull in a specific commit through composer by specifying the branch and commit instead of a version:
> 
> ```json
> "require": {
>     "laravel/telescope": "2.0#1749dae1"
> }
> ```
> 
> This will pull branch `2.0` at commit `1749dae1` (most recent).
> 
> The recent changes since the last version can be found here: [v2.0.4...2.0](https://github.com/laravel/telescope/compare/v2.0.4...2.0).

You need to add `dev-` for this to work:
```json
"require": {
    "laravel/telescope": "dev-2.0#1749dae1"
}
```

Issue #412, opened 2018-11-18
Request is not recorded when using ShouldBroadcastNow


rolandstarke:
When I trigger the following event the recording stops for this request.

```php
class Test implements ShouldBroadcastNow
{
//...

class BroadcastTestController extends Controller
{
    public function index(Request $request)
    {
        event(new Test());
        //this request is not recorded
//...
```

I think its because `ShouldBroadcastNow` triggers a `JobProcessed` event. and then the recording is stopped prematurely:

https://github.com/laravel/telescope/blob/81be58678d7d3a966651f2ac784fbdf8724b34b5/src/ListensForStorageOpportunities.php#L90

Related issue:  https://github.com/laravel/telescope/issues/201

paras-malhotra:
This seems to be a bug. @themsaid is there any reason why we need to call `stopRecording` and `startRecording`?

themsaid:
I can't replicate it, when I run a request that fires an event both the request and the event are recorded.

rolandstarke:
I will recheck this evening did you check with `ShouldBroadcastNow`?

themsaid:
I was able to replicate finally and fixed it. thanks :)

PR #623, opened 2019-05-03
fix events with ShouldBroadcastNow

josiasmontag:
[This fix](https://github.com/laravel/telescope/commit/9dfb1b08d3cf30adcfd44c79b5aaaa1f2fe34379) did not completely solve #412.

After dispatching an event with `ShouldBroadcastNow` the following entries are not correctly associated with the current request. 

This PR solves this and disables the premature storing if an event with the sync driver was emitted.

### DIFF
--- a/src/ListensForStorageOpportunities.php
+++ b/src/ListensForStorageOpportunities.php
@@ -83,12 +83,9 @@ trait ListensForStorageOpportunities
     {
         array_pop(static::$processingJobs);
 
-        if (empty(static::$processingJobs)) {
+        if (empty(static::$processingJobs) && $event->connectionName !== 'sync') {
             static::store($app[EntriesRepository::class]);
-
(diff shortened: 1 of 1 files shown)


themsaid:
@josiasmontag I'm not sure what's broken and this PR is trying to fix, what are the steps to re-produce?

josiasmontag:
```php
class TestEvent implements ShouldBroadcastNow
{
}
```


```php
class BroadcastTestController extends Controller
{
    public function index(Request $request)
    {
        DB::table('users')->get(); // recoded correctly
        event(new TestEvent());
        DB::table('users')->get() // recorded, but not associated to the current request
    }
}
```

The second query gets recorded but it does not show up in the "Queries" Tab on the Request detail page in Telescope. This PR solves it and the second query is correctly assigned to the current request.

PR #634, opened 2019-05-16
Env for slow limit on query watcher

jasonmccallister:
Adds an environment, with a default of 100, to the query watcher to easily override the slow limit on `telescope.php` config file. 

Not a breaking change.


taylorotwell:
Since you own this configuration file you can do this to your own file if needed.

PR #637, opened 2019-05-25
Fix empty telescope path config in frontend

mmachatschek:
This PR fixes the frontend, which would break if the `config('telescope.path')` would be set to either empty or `'/'`.

Additionally, the configuration for `config('telescope.path')` and `config('telescope.domain')` now supports the possibility to overwrite the values via the .env file.

There are no frontend tests, so unfortunately I can not provide any tests. I click tested as good as I could. 

**Benefits**

For environments (local and production) where you have a custom subdomain for telescope e.g. `telescope.example.local` you can now omit the prefix of the telescope routes by setting the .env variable option `TELESCOPE_PATH` to either empty or `"/"`

This makes telescope available via the following route `http://telescope.example.local/` instead of `http://telescope.example.local/telescope`


taylorotwell:
@themsaid can you review this?

themsaid:
thanks :)

mmachatschek:
@themsaid You're welcome

Issue #571, opened 2019-03-01
Search requests by status


oad-dwarf:
A common usecase for us is to investigate requests with particular HTTP status codes, primarily 422 and 500.

So it would be very nice if Telescope automatically added tags so the status is searchable, for example  `Status:422`.

themsaid:
I think tagging with the response status is not a bad idea, feel free to open a PR.

oad-dwarf:
Thanks @themsaid, but apparently I don't have permission to push a branch, so I cannot open a PR.

But the code is simple:

```patch
diff --git a/src/Telescope.php b/src/Telescope.php
index 2d89c8d..af8e6a2 100644
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -434,6 +434,7 @@ class Telescope
      */
     public static function recordRequest(IncomingEntry $entry)
     {
+        $entry->tags[] = 'Status:' . $entry->content['response_status'];
         static::record(EntryType::REQUEST, $entry);
     }


```

oad-dwarf:
@themsaid The code snippet works, I have tested it locally 👍

wotta:
What is the status on this issue?
I would love to see this functionality in Telescope.

So I opened a PR with this functionality in it.
See PR #641.

mchaabane:
You can add this tag in your TelescopeServiceProvider class : register() 

```
        // Add Telescope tag status
        Telescope::tag(function (IncomingEntry $entry) {
            if ($entry->type === 'request') {
                return ['status:'.$entry->content['response_status']];
            }
            return [];
        });
```

PR #641, opened 2019-05-29
Search requests by status

wotta:
Opened this PR for issue: #571 

If this is not the time to merge it please let me know what should be done for this to be added. 🙂

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -435,7 +435,7 @@ class Telescope
     public static function recordRequest(IncomingEntry $entry)
     {
         $entry->tags[] = 'Status:'.$entry->content['response_status'];
-        
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
To me tags should be fairly unique. Also, aren't you Abel to add this into your own tagging logic if you want it?

wotta:
Heej @taylorotwell, 

Thanks for the reply :) 
And unfortunately, I couldn't find a way to do this.

Maybe I didn't search enough but I couldn't find any documentation on how to do this.
Even on the documentation, I didn't see anything about adding custom tags to search on.

If I didn't search good enough and you know there is documentation that explains this would you be willing to share the documentation so I can look at it and implement this in my own project?

PR #642, opened 2019-05-29
Add app_name in dashboard

tomswinkels:
We have a few project and every project has telescope. We have place the telescope monitor on a screen with a switch between all the telescopes but we now can't see witch telescope it is.

We think this is a good solution.


taylorotwell:
You're not displaying the name in this PR though?

tomswinkels:
Whoops, you right..!

https://github.com/laravel/telescope/pull/643

PR #643, opened 2019-05-29
Add app_name in dashboard

tomswinkels:
We have a few project and every project has telescope. We have place the telescope monitor on a screen with a switch between all the telescopes but we now can't see witch telescope it is.

We think this is a good solution.

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -7,7 +7,7 @@
     <meta name="csrf-token" content="{{ csrf_token() }}">
     <link rel="shortcut icon" href="{{ asset('/vendor/telescope/favicon.ico') }}">
 
-    <title>Telescope</title>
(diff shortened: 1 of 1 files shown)


browner12:
Environment variables should not be directly accessed in views. Please use a config variable instead.

tomswinkels:
@browner12 Taylor always say not to much config variables?

browner12:
If an application uses configuration caching, all calls to `env()` will return `null`.

https://laravel.com/docs/5.8/configuration#configuration-caching

tomswinkels:
Thnx @browner12. What is the best place? config/telescope.php?

browner12:
it already exists in `config/app.php`.

`config('app.name')`.

tomswinkels:
You right. I have make a change in the pr.

PR #644, opened 2019-05-30
Add ignore request methods functionality

tanmuhittin:
This is a minor feature which adds the functionality to ignore requests with defined methods.

This is very useful if the application makes a lot of get requests in every page view and you are interested in inserts, deletes and updates. I needed this functionality in a Laravel Nova project where lots of get requests were being logged in every page view.


taylorotwell:
Can you not already register your own Telescope filter?

PR #656, opened 2019-06-26
Enable chunking on delete to prevent errors on large result sets

### LINKED ISSUE #536: Pruning very large table
Once the telescope_entries table gets large enough, prune commands start to fail. This ends up being due to just the amount of time it takes to run the prune command, but shouldn't just not prune things.

Steps to reproduce:
1. Use telescope either without running pruning, or just be using the app enough to generate thousands of records per day (maybe more).
2. Get the table into multiple millions of records
3. Watch the prune command timeout.

I am guessing that perhaps chunking the deletes would solve this entirely, but as it doesn't chunk on this or the clear() method, both seem to fail once the table gets big enough.

ragingdave:
This is to fix #536 

Actual metrics (including better math than napkin math):
(03:54 PM) ./artisan telescope:prune
2651547 entries pruned.
(04:04 PM)

### DIFF
--- a/src/Storage/DatabaseEntriesRepository.php
+++ b/src/Storage/DatabaseEntriesRepository.php
@@ -321,9 +321,17 @@ class DatabaseEntriesRepository implements Contract, ClearableRepository, Prunab
      */
     public function prune(DateTimeInterface $before)
     {
-        return $this->table('telescope_entries')
-                ->where('created_at', '<', $before)
-                ->delete();
+        $query = $this->table('telescope_entries')
+                ->where('created_at', '<', $before);
+
+         $deleted = 0;
+
+         do {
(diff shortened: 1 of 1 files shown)


themsaid:
@ragingdave tests are failing after your PR.

ragingdave:
I saw that they were but I'm pretty sure that they are failing due to missing code on master from PR #653. Would you like me to change the base branch and resubmit based on 2.0 code?

ragingdave:
Specifically it looks like it's been failing an has failed the same way on my PR as a result of missing the fix that fixes the tests: https://travis-ci.org/laravel/telescope/jobs/545905562 as an example

taylorotwell:
Please send to 2.0 branch since this is not a breaking change.

PR #666, opened 2019-07-18
Chunk storage of exceptions and tags

Namoshek:
This PR is a follow-up of #615 which was a fix for SQL Server issues with large transactions where a lot of Telescope entries were created (explanation of issue in #614).

In the previous PR, I only fixed `DatabaseEntriesRepository::store($entries)` as it resolved my issues. Unfortunately, with some changes and new features a new issue came up with storage of tags in `storeTags($results)`. The solution is analogous to the previous PR and I also updated `storeExceptions($exceptions)` while being at it.

On a site note: I extracted some counting logic of `storeExceptions($exceptions)` into its own method to simplify the logic a bit. If this change is not desired, I will remove it. It is fully backwards-compatible though.

### DIFF
--- a/src/Storage/DatabaseEntriesRepository.php
+++ b/src/Storage/DatabaseEntriesRepository.php
@@ -108,6 +108,20 @@ class DatabaseEntriesRepository implements Contract, ClearableRepository, Prunab
     }
 
     /**
+     * Counts the occurences of an exception.
+     *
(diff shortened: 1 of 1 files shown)


taylorotwell:
Was totally broken becuase of missing import. Please make sure you actually run the tests included with the code base before sending us a PR. That is why they are there. Thanks.

Namoshek:
I see why, I'm sorry. Thanks for merging though.

PR #670, opened 2019-07-25
wrap service provider boot code in telescope.enabled conditional

mattquest:
Currently, disabling telescope through the `TELESCOPE_ENABLED` environment variable will prevent `Telescope::start` from running but does not prevent telescope from registering routes, migrations, views etc. in the `TelescopeServiceProvider->boot` method.  

This can cause unexpected errors and problems in situations like a testing environment where telescope is disabled and the migrations are not expected to be registered.

### DIFF
--- a/src/TelescopeServiceProvider.php
+++ b/src/TelescopeServiceProvider.php
@@ -18,18 +18,20 @@ class TelescopeServiceProvider extends ServiceProvider
      */
     public function boot()
     {
-        Route::middlewareGroup('telescope', config('telescope.middleware', []));
+        if (config('telescope.enabled') === true) {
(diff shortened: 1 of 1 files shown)


Namoshek:
This PR actually broke some Dusk tests of mine because the Telescope route of a link in the menu was not registered.

Disabling a package entirely seems very drastic to me. I would suggest reverting this and letting affected users registering the `TelescopeServiceProvider` only in non-testing environments:
```php
// in AppServiceProvider::register()
if ($this->app->environment('testing') === false) {
    $this->app->register(\Laravel\Telescope\TelescopeServiceProvider::class);
}
```

PR #679, opened 2019-08-12
Add view collector

### LINKED ISSUE #677: Views/Templates watcher
I think View/Template information could be very useful. Do you have plans for this, or are you okay with making a PR for this?

Wouldn't have to be much, but just the views that are rendered (eg view name / actual filename / maybe parameters that are passed or something.

barryvdh:
This adds a Watcher for the Views/templates. 
The RequestWatcher logs the latest view, but not any other views/subviews which might be useful.
The `extractDataFromView` could be re-used, but I think that would generate a lot of data / cost in CPU.
Icon is the same as Model, not sure where the SVG's are coming from to pick one.
Fixes #677
(Assets are not compiled because of potential merge conflicts)

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -152,5 +152,7 @@ return [
         ],
 
         Watchers\ScheduleWatcher::class => env('TELESCOPE_SCHEDULE_WATCHER', true),
+
+        Watchers\ViewWatcher::class => env('TELESCOPE_VIEW_WATCHER', true),
     ],
 ];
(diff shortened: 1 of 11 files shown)


taylorotwell:
Can you screenshot what it looks like?

barryvdh:
I've updated the path to strip the base path, and filters some data that is always present.

Index:
image
Details:
image

Request tab:
image

taylorotwell:
The icons are Zondicons it looks like: https://www.zondicons.com/icons.html

taylorotwell:
Maybe `show-sidebar` icon?

barryvdh:
Sure, updated it.

barryvdh:
Thanks? Don't forget this still needs a compilation of assets.

PR #683, opened 2019-08-14
Optimize dump

### LINKED ISSUE #611: Dump is not working
- Telescope Version: 2.0.4
- Laravel Version: 5.8.9
- PHP Version: 7.2.16

### Description:

`dump()` is not working. I'm not seeing anything in the dump section in Telescope.

### Steps To Reproduce:

See https://github.com/laravel/telescope/issues/487#issuecomment-479425001

Commenting the return statement in the DumpWatcher class seems to do the trick, but it's only a work-around.

Thank you

### LINKED ISSUE #682: [Dump] DumpCollector can be slow with large amount of data
The DumpCollector works by continuously loading the Dump entries. There are 50 dump entries loaded on each request by default, which are refreshed every 2 seconds. 
For the other collectors, there is a `?take=1` parameter which is compared to the last entry. Perhaps that should be done with Dump also?

And besides that, when dumping large object, Dump can become very heavy. Try generating 50 dumps of `app()` and loading the page (not very real-life, but demonstrates the problem -> generating 10+MB of HTML and refreshing that every 2 seconds.

Would the Dump screen perhaps make more sense as a normal preview page?

barryvdh:
This shuld fix #682 and partially fix #611 (or at least give the user some information about it).
Addition could be to specify a cache driver in config/telescope, so that the Telescope driver could be different (eg. if you want to work with the array driver locally).

This shows an error when something goes wrong (eg. the wrong cache driver in this instance).
It also changes the loading of entries to check 1 entry and only load the rest when something changed (just like the rest of the Index screens).

### DIFF
--- a/resources/js/screens/dumps/index.vue
+++ b/resources/js/screens/dumps/index.vue
@@ -11,7 +11,7 @@
                 ready: false,
                 error: null,
                 newEntriesTimeout: null,
-                newEntriesTimeoutInSeconds: 2000,
+                newEntriesTimer: 2000,
                 recordingStatus: 'enabled'
             };
         },
@@ -43,9 +43,7 @@
                     this.entries = response.data.entries;
                     this.recordingStatus = response.data.status;
 
-                    this.newEntriesTimeout = setTimeout(() => {
-                        this.loadEntries();
-                    }, this.newEntriesTimeoutInSeconds);
+                    this.checkForNewEntries();
                 }).catch(error => {
                     if (error.response && error.response.data.message) {
                         this.error = error.response.data.message;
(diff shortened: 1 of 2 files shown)


driesvints:
Are there breaking changes here? If not then shouldn't this be sent to 2.0 branch instead?

barryvdh:
To be honest, I'm not really sure which branch is correct. I saw the latest development happen in the master branch, and as far as I can tell, none of them are breaking changes. So not sure why this branch is 3.0 instead of 2.1, or why there is a seperate branch. Looking at it now, they are only fixes / small changes: https://github.com/laravel/telescope/compare/master

driesvints:
@barryvdh 3.0 hasn't been released yet. All our repos follow the same branching system where the master is always the next major version. Backwards compatible new features can always be sent to the current release branch: https://laravel.com/docs/5.8/contributions#which-branch

barryvdh:
Okay, then I think you can merge all commits from master into 2.0 and I'll update my open PR's to 2.0 after that.

driesvints:
@barryvdh I got a bit stuck with cherry-picking so feel free to backport yourself if you want this in the next 2.x release.

barryvdh:
Okay, which one do you not want to include on 2.0? Looking through the commits, the only one that might be unneccesary is your dev-master update, but that doesn't matter on branches that are not dev-master, right?

driesvints:
@barryvdh I wanted to backport but the cherry-picking just failed for me. So didn't want to lose too much time on this. Just feel free to send this in yourself.

barryvdh:
@driesvints so https://github.com/laravel/telescope/pull/686 ?

PR #685, opened 2019-08-15
Log failed requests

barryvdh:
Check if an IncomingEntry is a failed (5xx) Request and log that. Useful to log request info with the Exceptions etc.


dbpolito:
Great PR but looks like you got a failing test.

barryvdh:
Yeah that was from a previous PR, which is fixed in https://github.com/laravel/telescope/pull/684 but the tests need to be re-run.

PR #686, opened 2019-08-15
Backport fixes to 2.0

barryvdh:
As fas as I can tell, none of these changes are breaking / big features, so can be merged to 2.0


driesvints:
@barryvdh seems like you're trying to merge in all commits from the master branch. You should try to only cherry-pick the commits from the feature. Can you try a new PR?

barryvdh:
I was indeed trying to merge all the commits from master, because they are all bugfixes and should probably be in 2.0? Or which one should I exclude? 
I can cherry-pick some, but not sure which you don't want.

driesvints:
@barryvdh I wasn't aware that you sent in more prs. I suggest that you sent in the same prs that you sent to the 3.0 branch also to the 2.0 branch.

barryvdh:
But that would make the history even more confusing, with diverted paths? And some other changes that were not mine.

But anyways, my changes aren't that important (mostly usability/small features) that they have to be in it NOW, so I'll just leave it as is.  Not sure what the timing of 3.0 is going to be.

driesvints:
Hey @barryvdh. I talked to Taylor about this and this seems good to be merged. Thanks for sending this in 👍

PR #689, opened 2019-08-16
Dump queued items when die'ing

barryvdh:
This is a potential fix for `dd($something)` which kills the script when the dump's are not persisted yet. It dumps the existing entries on screen, just like a regular `dd()`.

Not sure what the impact would be, but usually it wouldn't change anything (only happens when you are watching the dumps + when you die with something in the queue)


taylorotwell:
No plans to merge this at this time.

Issue #691, opened 2019-08-17
[Testing] How to get IncomingRequest for testing purposes?


stancl:
I want to test that a request has specific tags. How do I do that?

barryvdh:
You could set the `afterRecording` perhaps? It doesn't give you the entry itself (not sure why not though), but you can get it from the entries queue (static property)

stancl:
On second thought, for my use case it might be better to just make a request and check the `telescope_entries_tags` table.

PR #692, opened 2019-08-17
[3.0] Change $tagUsing to array

### LINKED ISSUE #673: [Feature Request] A way to add tags as a package
Right now, `Telescope::tag()` sets the callback that returns the user-specified tags.

https://github.com/laravel/telescope/blob/d7a3ca4338b1a1527b8d19a5d62015a468f8ca56/src/Telescope.php#L521-L532

This doesn't allow packages to easily set tags, because they would override the user-specified tags.

An example of when a package would want to set a tag is multi-tenancy. I'd like my package to set a tag with the tenant id for each request.

It might be possible to set tags from a package using something like this:

```php
$original_callback = Telescope::tagUsing;

Telescope::tag(function (IncomingEntry $entry) {
    if ($entry->type === 'request') {
        $tags = ['status:'.$entry->content['response_status']];
    } else {
        $tags = [];
    }

    return array_merge($original_callback($entry), $tags);
});
```

but it would be nice to have a less hacky way of setting multiple tag-creating callbacks.

stancl:
Resolves #673.

There can now be multiple tag-setting callbacks. This is useful if a package wants to set tags without overriding tags set by the user.

This could be done in v2 in an ugly way:
```php
$originalCallback = Telescope::$tagUsing;

Telescope::tag(function ($entry) use($originalCallback) {
    $originalTags = [];
    if (! is_null($originalCallback) {
        $originalTags = $originalCallback($entry);
    } 

    $tags = // custom logic

    return array_merge($originalTags, $tags);
});
```

And it would work only with two tags, unless all callbacks did this logic.

This PR makes this much nicer.

I would have added tests, but not sure how to do that (related to #691).

<sub>(sorry for the slightly messy PR, made this from 2.0)</sub>

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "3.0-dev"
+            "dev-master": "2.0-dev"
         },
         "laravel": {
             "providers": [
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -248,7 +248,7 @@ class Telescope
         }
 
         $entry->type($type)->tags(array_map(function ($tagCallback) use ($entry) {
-            $tagCallback($entry);
+            return $tagCallback($entry);
         }, static::$tagUsing));
 
         try {
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -42,11 +42,11 @@ class Telescope
     public static $afterRecordingHook;
 
     /**
-     * The callback that adds tags to the record.
+     * The callbacks that add tags to the record.
      *
-     * @var \Closure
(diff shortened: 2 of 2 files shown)


stancl:
Perhaps the `$afterRecordingHook` should also be changed into an array, for consistency?

taylorotwell:
There are a lot of other commits on here other than what you describe in your PR.

stancl:
Those commits come from the 2.0 branch so you would have merged them in
anyway.

I'll open a new PR with just the two commits of mine.

Dne ne 18. srp 2019 16:01 uživatel Taylor Otwell <notifications@github.com>
napsal:

PR #698, opened 2019-08-19
Timing screen

### LINKED ISSUE #674: [Feature Request] Measuring Tool / Timeline
Is there some way to get a timeline for the requests? I.e. I would like a function like:
start_measure($id), end_measure($id) and then I would get those timings logged in telescope. 
Similar to the way debugbar and clockwork provide timelines. 
(This feature is actually the only reason why I still have to use clockwork alongside Telescope)

barryvdh:
This is a Draft Pull Request to check if this is worth including, before too much time is spent on this.

- Adds timing information to entries
- Shows timing information
- Ability to start/stop timers

Fixes #674 

Example:
Screen Shot 2019-08-15 at 22 19 56

TODO
 - [ ] Fix layout to make it useful
 - [ ] Think about how to add timing (eg. sections? perhaps use https://symfony.com/doc/current/components/stopwatch.html ? )
 - [ ] Record the timing data together with the IncomingEntry, or create 1 timing entry with just all the events?
 - [ ] Add timing info for all appropriate watchers.
 - [ ] Add framework timing info (boot etc)

cc @christophmayrhofer

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "3.0-dev"
         },
         "laravel": {
             "providers": [
--- a/resources/js/routes.js
+++ b/resources/js/routes.js
@@ -180,4 +180,10 @@ export default [
         name: 'gates',
         component: require('./screens/gates/index').default,
     },
+
+    {
+        path: '/timing/:id',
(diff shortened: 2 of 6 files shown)


barryvdh:
@taylorotwell Do you think this is worth considering?

taylorotwell:

taylorotwell:
Better to use something more robust like Blackfire, imo.

PR #699, opened 2019-08-19
[2.0] Fall back to database.default when telescope.storage.database.connection is null

stancl:
This makes Telescope more flexible when dynamically changing the default DB connection, such as with multi-tenant setups.


stancl:
Actually, it doesn't make Telescope more flexible, because the value is read from config once.

PR #706, opened 2019-08-22
Format SQL in watcher

barryvdh:
This formats the SQL in PHP (instead of Javascript) so that they can be shown in the preview window (imho that makes it a lot more clear when looking at them, but might be preference, hence the config options).
It stores the familyHash to be able to check for duplicates (which isn't implemented anywhere yet, but would perhaps be useful).

The view of the Javascript should be the same, with/without pre-formatted SQL.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -136,7 +136,6 @@ return [
             'enabled' => env('TELESCOPE_QUERY_WATCHER', true),
             'ignore_packages' => true,
             'slow' => 100,
-            'format_sql' => true,
         ],
 
         Watchers\RedisWatcher::class => env('TELESCOPE_REDIS_WATCHER', true),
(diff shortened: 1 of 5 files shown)


barryvdh:
Added summary of the queries (total, duplicated and total duration):
Screen Shot 2019-08-22 at 21 55 41

taylorotwell:
It's a little unclear to me what the "format_sql" configuration option does. I think it will also be unclear to end users.

barryvdh:
So how would you call it? Or would you do it as default?

barryvdh:
It's replacing the `?` parameters with the actual bindings, so you can see the query as if it was prepared. So instead of `id = ?` you get `id = 3`, so basically what the javascript was doing now on details, but now directly.

barryvdh:
Not sure, because it wasn't done before I thought maybe that was on purpose.

barryvdh:
I've removed the option, just always do it now.

barryvdh:
Fixed+added a test.

PR #710, opened 2019-08-23
Mark exceptions as resolved

barryvdh:
This allows to mark exceptions as resolved, and show warnings when not resolved.

image

image

If an Exception occurs again, it's shown as not-resolved again (because the last entry is shown, which is good).

### DIFF
--- a/resources/js/screens/exceptions/index.vue
+++ b/resources/js/screens/exceptions/index.vue
@@ -6,9 +6,10 @@
     <index-screen title="Exceptions" resource="exceptions">
         <tr slot="table-header">
             <th scope="col" v-if="!$route.query.family_hash">Type</th>
-            <th scope="col" v-if="!$route.query.family_hash && !$route.query.tag">Occurrences</th>
+            <th scope="col" v-if="!$route.query.family_hash && !$route.query.tag">#</th>
             <th scope="col" v-if="$route.query.family_hash">Message</th>
             <th scope="col">Happened</th>
+            <th scope="col">Resolved</th>
             <th scope="col"></th>
         </tr>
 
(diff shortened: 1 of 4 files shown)


taylorotwell:
The button looks slightly too low. It's not vertically aligned with the text on the left.

barryvdh:
Is this better?

image

taylorotwell:
Hmm, it looks more aligned but now the top padding on the button appears much less than the bottom padding. Sorry I am picky about these things 😬

barryvdh:
But isn't that exactly the same as the Auth tag? Tried using the built in styling classes.

barryvdh:
Do you mean inside the button? Because the padding above the lowercase letters it the same top and bottom. Only the uppercase 'R' and 'l' are closer to the top. So not really sure how it should be aligned otherwise:
image

GertjanRoke:
Maybe the `vertical-align` of the `.badge` class needs to be changed to `text-align: center;` that fixed it for me.

taylorotwell:
@GertjanRoke can you screenshot that?

GertjanRoke:
image
The right one is with `vertical-align: middle;`

barryvdh:
I can change that, but that would be benifitial for all `.badge` items, so not sure if that's related to this PR.

GertjanRoke:
@taylorotwell I thought I would make a better screenshot for you by rebuilding what @barryvdh had in his first screenshot.

So this one is with `vertical-align: bottom`:  

image

And this one is with `vertical-align: middle`:  

image

And this is with the current styling:  

image

Hope this makes it easier to decide what the best option is.

taylorotwell:
I'll just go as-is for now.

barryvdh:
Thanks, should I submit a PR to rebuild assets and perhaps tag a new release after that, to prepare for 6.0?

PR #721, opened 2019-09-05
Update vue-json-pretty

Dannybevers:
Updating vue-json-pretty for telescope to ^1.6.2


tim-hanssen:
This fixes the payload views making it a valid json so we can use the output in debugging tools without replacing every key.

DavydeVries:
Before:
Schermafbeelding 2019-09-05 om 12 00 54

After:
Schermafbeelding 2019-09-05 om 12 01 03

Issue #726, opened 2019-09-06
null displayed as empty string


tillkruss:
- Telescope Version: 2.0
- Laravel Version: 6.0
- PHP Version: 7.3

### Description:

I'm executing an increment on a pivot model:

```php
$participant->increment('unread', 1, ['archived_at' => null]);
```

Telescope shows this query:

```sql
update
  "conversations_companies_map"
set
  "unread" = "unread" + 1,
  "archived_at" = '',
  "updated_at" = '2019-09-06 22:16:15'
where
  "id" = 2386
```

Could it be possible that Telescope is displaying `null` as `''` by chance?

barryvdh:
Which version exactly? You say 2.0 but 2.1 is the latest, which includes https://github.com/laravel/telescope/pull/706

tillkruss:
Sorry, 2.1

barryvdh:
See https://github.com/laravel/telescope/pull/727

tillkruss:
You’re the best! ❤️

PR #727, opened 2019-09-07
Check for null values in query bindings

barryvdh:
Checks for `null` values in queryies. Fixes https://github.com/laravel/telescope/issues/726


barryvdh:
Updated to tests other values and also test the named parameters.

Issue #594, opened 2019-03-21
Publish assets with Composer hook


driesvints:
Like https://github.com/laravel/horizon/issues/554 it should also be possible for Telescope to publish updated assets after a composer dependency update. Telescope will need to add its own assets command for this.

barryvdh:
An alternative to this could be:
 - Render the assets inline (downside: no caching, loaded each request but not so big an issue because it's loaded only on the first request)
 - Pass the assets through the API (eg `/telescope/assets/script.js` -> AssetController@getScriptJs`, which could set a TTL for browser caching)

PR #729, opened 2019-09-09
Show warning when manifest is outdated

barryvdh:
This adds a check to the base view to show a warning when the manifest file in the public folder is different from the one in the vendor dir. This prevents broken components or missing updates.

image

The exception is also updated to make it more clear.
Related: #594

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -709,7 +709,7 @@ class Telescope
      * @throws \Illuminate\Contracts\Filesystem\FileNotFoundException
      * @return bool
      */
-    public static function assetsUpdated()
+    public static function assetsAreCurrent()
     {
         $publishedPath = public_path('vendor/telescope/mix-manifest.json');
 
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -7,6 +7,7 @@ use Exception;
 use Throwable;
 use Illuminate\Support\Str;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\Facades\File;
 use Illuminate\Log\Events\MessageLogged;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Laravel\Telescope\Contracts\EntriesRepository;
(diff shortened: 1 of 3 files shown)


taylorotwell:
Method name doesn't make sense to me. Make it something like "assetsAreCurrent" or "assetsOutOfDate"

barryvdh:
Renamed to `assetsAreCurrent`

tomswinkels:
Why not autoupdate the assets?

barryvdh:
The browser usually runs with different permissions and there is no hook in composer to define in this package (only in the root package)

src/Telescope.php
```diff
@@ -701,4 +702,21 @@ public static function ignoreMigrations()
...
+     *
+     * @throws \Illuminate\Contracts\Filesystem\FileNotFoundException
+     * @return bool
+     */
+    public static function assetsAreCurrent()
+    {
+        $publishedPath = public_path('vendor/telescope/mix-manifest.json');
+
+        if (! File::exists($publishedPath)) {
+            throw new \RuntimeException('The Telescope assets are not published. Please run: php artisan telescope:publish');
```
stancl: Could this be a custom exception (like `TelescopeAssetsNotPublishedException`) that implements `ProvidesSolution`?

barryvdh: Technically it could yeah

barryvdh: Not sure what isn't clear about this error and the solution though :P

stancl: There could be a button to run `Artisan::call('telescope:publish')`, similarly to the migrate button when a table is not found.

tomswinkels:
@barryvdh can you make a PR for Laravel Horizon?

barryvdh:
I don't think @driesvints was convinced yet.

tomswinkels:
@barryvdh but i think it is better to do the same in Horizon and Telescope :) @driesvints

driesvints:
Sorry, I'm still opposed to do anything about this in the UI. But feel free to make the pr to see if Taylor will accept it.

PR #737, opened 2019-09-16
Use time instead of duplicated to show global queries duration

cmizzi:
This patch is pretty simple. The duration shown on the screen below is wrong because it shows the number of duplicated queries instead of sum of queries duration.

image


barryvdh:
Whoops, missed that. Thanks!

PR #752, opened 2019-10-17
Add the Nova API to ignore paths array

WyattCast44:
Hello, 

Would you be open to by default ignoring the nova API by default? You can do this individually by project via the config file but the Horizon API is ignored by default so I figured it's worth asking.

The Nova API creates a lot of requests that don't provide much value (in my case), if others get value from those request logs then disregard.

Cheers, Wyatt!


bolechen:
can make this configurable? i want to debug nova api.

PR #771, opened 2019-11-13
Configure telescope to use bootstrap's container-fluid class

hendeavors:
Telescope by default uses bootstrap's `container` class. This can sometimes cause items to flow out of  the table:

image
 
This update allows telescope to use bootstrap's `container-fluid` class by simply using `Telescope::fluid();` .


taylorotwell:
I don't think this should be configurable. I think we should use whatever is best by default.

PR #772, opened 2019-11-14
[2.x] Use container-fluid for layout

hendeavors:
I think its best to use bootstrap's `container-fluid` class to prevent the below:
image


themsaid:
Thanks but there's no plan to merge this at the moment :)

Issue #775, opened 2019-11-19
Request response missing in some cases and replaced with a generic "HTML Response"


damms005:
Some request have response simply set as "HTML Response".

This can be made better

PR #777, opened 2019-11-19
Improves logged request response

damms005:
Improves on logged responses. Addresses #775


taylorotwell:
No plans to change this at the moment.

damms005:
Does it count if I add that this change enables the ability to track some kind of (likely api) fails:

`return response( 'some irregular stuff happened', 500 )`

Instead of having to first `Log.error( 'some irregular stuff happened' )` then `return response( 'some irregular stuff happened' )`?

PR #783, opened 2019-12-05
Update to 2.1 alias

barryvdh:
The branch is 2.0, but the last version is 2.1 so I think this needs to be updated?
Otherwise you have to require `2.0-dev` which is a bit weird.

Perhaps a new tag would also be possible?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -35,7 +35,7 @@
     },
     "extra": {
         "branch-alias": {
-            "dev-master": "2.0-dev"
+            "dev-master": "2.1-dev"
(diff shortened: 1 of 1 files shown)


taylorotwell:
Eh, I don't think this actually does anything now that there is a 2.0 branch?

barryvdh:
So should the 2.0 branch be renamed to 2.1? Right now the dev version is lower than the tagged on it seems:

Screen Shot 2019-12-06 at 00 04 26

PR #810, opened 2020-01-15
[2.x] Avoiding to create guard

dkulyk:
Sometimes guard is created when session not initialized yet.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -256,7 +256,7 @@ class Telescope
         );
 
         try {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why does this matter?

dkulyk:
QueryWatcher try to make guard when session query executed and the guard not ready yet.
```
Request
 -> Make guard
  -> Make session store(stored to container)
   -> Query
    -> QueryWatcher
     -> Make guard(First guard not stored yet)
      ...
```

PR #821, opened 2020-02-01
[2.x] Use Illuminate\Foundation\Application instead of contract

stancl:
The stub service provider uses `$this->app->isLocal()` and that method exists only on `Illuminate\Foundation\Application`, not the contract.

Screenshot from 2020-02-01 13-07-55


driesvints:
@stancl can we maybe change the call to `$this->app->environment('local')` instead? This way the property override isn't needed.

Overriding a property with a new type doesn't enforces it.

stancl:
Sure. I think this was the only case of depending on the concrete's public API rather than the abstract's API, so if that solution will work then sure, this is a better way.

I will change this later today.

taylorotwell:
Feel free to re-open when the change is made.

PR #857, opened 2020-04-01
Switch to brick/math

tomcastleman:
Package `moontoast/math` is abandoned. This PR switched to suggested replacement `brick/math`.


tomcastleman:
Looks like this will never be compatible with Laravel <7.0 due to older `ramsey/uuid` dependency.

Issue #862, opened 2020-04-05
AfterRecordingHook doesn't pass the IncomingEntry object


MatanYadaev:
For some reason the `AfterRecordingHook` passes the `Telescope` object instead of the `IncomingEntry` object, is that a reason for this behavior? 

For my use-case, I would prefer receiving the `IncomingEntry` itself, to be able to do some logic on it.

https://github.com/laravel/telescope/blob/b6f903533d054030f0c24def89fa7c57fd33c3ed/src/Telescope.php#L272

Origin: https://github.com/laravel/telescope/pull/462/files#r403687515

driesvints:
No idea really. Feel free to PR. Think this might have to go into the next version branch.

MatanYadaev:
I want to PR. Should I replace this parameter or add besides it?

driesvints:
Replace I think

driesvints:
Closing as it turns out that this is intentional.

PR #867, opened 2020-04-12
[3.x] Pass IncomingEntry in AfterRecordingHook

MatanYadaev:
As discussed in #862, the wrong parameter is passed in this event, it should be the model itself, `IncomingEntry`.
I fixed it and fixed the test that used the previous implementation.

BTW, do you think this hook should be documented in the official [documentation](https://laravel.com/docs/7.x/telescope)?

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -270,7 +270,7 @@ class Telescope
             }
 
             if (static::$afterRecordingHook) {
-                call_user_func(static::$afterRecordingHook, new static);
+                call_user_func(static::$afterRecordingHook, $entry);
             }
         });
     }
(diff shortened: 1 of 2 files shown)


taylorotwell:
This is intentional:

image

MatanYadaev:
@taylorotwell Why does it intentional? What the advantage of using `$telescope::` instead of `Telescope::`?
Isn't the convention is to provide the entity instance as the first parameter in any before/after events?

adriaanzon:
You can use `Arr::last($telescope::$entriesQueue)` but it will not reliably return the entry that has been recorded before running the `afterRecordingHook`.

For example, in production, some entries get filtered out by the `Telescope::filter` callback, but the `afterRecordingHook` still gets called. If you would then simply use `Arr::last($telescope::$entriesQueue)` to retrieve the telescope entry e.g. for logging exceptions to Slack, an exception may be logged multiple times in a request, because the last item in `$telescope::$entriesQueue` may still be the telescope entry for the exception while the  `afterRecordingHook` is being called.

barryvdh:
In any case the entry should added as extra parameter, not replace the existing one, to avoid BC.

barryvdh:
Here as extra argument: https://github.com/laravel/telescope/pull/893

PR #870, opened 2020-04-14
[3.x] Feature: Copy to Clipboard

diogogomeswww:
A lot of times I need to copy data from telescope, for example, copying a SQL query, request payload,...

This PR will add a "Copy to Clipboard" icon and action to some elements in the UI:
- Request payload, headers, session and response
- Command arguments and options
- SQL Query
- ...

Screenshot:
image

Questions:
- From where should we copy the copy-2-clipboard svg icon?
- Right now when clicking it will hide the icon and show "copied..." for 2 seconds and show the icon again. Please let me know if you have a better idea.


public/mix-manifest.json
```diff
@@ -1,5 +1,5 @@
 {
-    "/app.js": "/app.js?id=58d232fb9ec88bf5e0e5",
-    "/app.css": "/app.css?id=326b997a579e007d2102",
-    "/app-dark.css": "/app-dark.css?id=5ebb5e33b7c425d6317c"
+    "/app.js": "/app.js?id=dc2c875341091f2cb128",
+    "/app.css": "/app.css?id=2f01ce9c781b2f50f6b1",
+    "/app-dark.css": "/app-dark.css?id=afbebebe35d8dec2afed"
```
driesvints, line 4: Please undo the compiling. We'll dot that ourselves once merged.

diogogomeswww: Done.

taylorotwell:
Doesn't work at all in Safari on my tests.

hrvoj3e:
@diogogomeswww will you try again or did you give up?
I seems to me like a great idea.
PR was rejected because of browser compatibility with Safari - right?
 - https://developer.mozilla.org/en-US/docs/Web/API/Clipboard#browser_compatibility
   - Safari seems supported now

diogogomeswww:
@hrvoj3e thank you, I thought it was a good idea too.
But it seems that the maintainers don't think so.

Issue #876, opened 2020-04-19
Queries with runtime over 1s are not summed up.


janolivermr:
- Telescope Version: 2.1.7
- Laravel Version: 6.18.8
- PHP Version: 7.4.4
- Database Driver & Version: Mysql

### Description:
Queries that run longer than 1 second are not summed in the request queries list.
This is likely due to the use of the `number_format` function, which adds thousands separators.
In the vue frontend, the reduce function uses a `parse_float` which can't handle the thousands separator.

### Steps To Reproduce:
Make a DB query that takes longer than 1 second inside a request.
View the request in Telescope and check the queries list. The long running query time will not be taken into account for the query time total.

driesvints:
Hey there,

Unfortunately we don't support this version anymore. Please check out [our support policy](https://laravel.com/docs/releases) on which versions we are currently supporting. Can you please try to upgrade to the latest version and see if your problem persists? We'll help you out and re-open this issue if so.

Thanks!

janolivermr:
Hi,

The code hasn't changed in the 3.x branch, it still uses the same `number_format` code in PHP and `parseFloat` in the JS front end function.
In my merge request ( #877 ) you mentioned the locale issue, however `parseFloat` understands only decimal points if I'm not mistaken.

driesvints:
@janolivermr let's see how the pr goes

PR #877, opened 2020-04-19
[3.x] Remove thousands separator in time records (#876)

janolivermr:
This overrides the default thousands separator in the PHP `number_format` function used to format the time entries. This allows requests that take over 1s/1000ms to be summed up correctly.

For a query with a running time of 1.2s, the current time entry in the database will be `1,200.00`. When this is passed to `parseFloat` in the front end it will be parsed as `1`.
With the change, the value will be stored as `1200.00` and correctly parsed as `1200`.


driesvints:
This won't work if your locale is set to something like `fr_FR` for example.

driesvints:
@janolivermr I misunderstood what was going on, thanks for the pr!

Issue #875, opened 2020-04-17
Ability to review Exception context data


crynobone:
### Current Behaviour

When an exception is thrown, only `message`, `location` and `stacktrace` can be reviewed from Telescope.

<img width="984" alt="Screenshot 2020-04-18 at 7 34 23 AM" src="https://user-images.githubusercontent.com/172966/79621817-58298a80-8147-11ea-8944-0a2b0ba708d9.png">

```
[2020-04-17 23:44:35] production.ERROR: The given data was invalid. {"method":"renewals.sync","error":{"vehicle.year_manufactured":["The selected vehicle.year manufactured is invalid."]},"exception":"[object] (Minions\\Exceptions\\ClientHasError(code: -32602): The given data was invalid. at /var/www/drivemark-platform/releases/17042020_2328133_841/vendor/katsana/minions/src/Client/Response.php:48)
[stacktrace]
#0 /var/www/drivemark-platform/releases/17042020_2328133_841/vendor/katsana/minions/src/Client/Project.php(63): Minions\\Client\\Response->validate()
```

```php
Log::error($this->getMessage(), [
    'method' => $this->getRequestMethod(),
    'error' => $this->getRpcErrorData(),
    'exception' => $this,
]);
```

### Change Request

It would be nice if we can have context information be available on Telescope so we can review this information.

driesvints:
What context data?

crynobone:
> What context data?

![Screenshot 2020-04-20 at 4 27 06 PM](https://user-images.githubusercontent.com/172966/79730725-d5115b80-8323-11ea-96d2-f4d8d9fefe0a.png)

driesvints:
Isn't this info available in `/telescope/logs` ?

crynobone:
> Isn't this info available in `/telescope/logs`?

Yes, but as the title said, it would be nice to have it available if it's also available with **Exception**

<img width="1580" alt="Screenshot 2020-04-20 at 5 36 57 PM" src="https://user-images.githubusercontent.com/172966/79737217-91bbea80-832d-11ea-8b01-5274b6ec61ae.png">

What do you think?

driesvints:
Yeah I agree 👍

PR #882, opened 2020-04-28
[3.x] Ability to review Exception context data

crynobone:
### Current behaviour

When developer use custom report and log an error manually using the following:

```php
<?php

namespace App\Exceptions;

class TestException extends \Exception
{
    public function report()
    {
        \Log::error($this->getMessage(), [
            'message' => 'Context message here',
            'path' => base_path(),
            'exception' => $this,
        ]);
    }
}
``` 

We should be able to see the custom context. However at the moment only `message`, `location` and `stacktrace` can be reviewed from Telescope.

Screenshot 2020-04-18 at 7 34 23 AM

### Excepted Behaviour

Exception without any custom context should remain the same as before.

image

Exception with custom context should be able to show the context.

image

Solves #875


taylorotwell:
Thanks

PR #885, opened 2020-05-05
[3.x] Add index to created_at

keyboardSmasher:
When the `telescope_entries` table gets very large, `telescope:prune` will fail miserably, actually bringing an AWS db instance to its knees. This is because `telescope:prune` is using `created_at` in a where clause when there's no index on that column.


barryvdh:
You might want to add the index the same way as the other indexes, but the idea looks fine to me ;)

Issue #742, opened 2019-09-26
Disable Gravatar or allow third-party services


RichardDern:
It would be nice to add a configuration entry to disable gravatar, if it doesn't already exists (I didn't see it in the docs)

Possible use-cases and concerns:

- Offline servers
- Privacy (I don't use Gravatar at all, so I don't want my email address to be sent to them with everything that implies in terms of security and privacy)

Another possibility would be to allow customising avatar's URL, so we can use third-party/self-hosted avatar-hosting services.

michael-schaefer-eu:
Hey +1 for the config switch.

I could find out that it sends only a hash of the email address, but anyway they get your IP and they do track the user!

```
gravatarUrl() {
    if (this.entry.content.user.email) {
        const md5 = require('md5')
        return 'https://www.gravatar.com/avatar/' + md5(this.entry.content.user.email.toLowerCase()) + '?s=200'
    }
}
```

derekmd:
@taylorotwell @driesvints This issue can be closed now `Telescope::avatar()` is available.

PR #890, opened 2020-05-10
[3.x] Allow avatar path customization

derekmd:
Covers: https://github.com/laravel/telescope/issues/742

### Features

✔️ Apps can disable Gravatar and stop telling its parent Automattic about user activity on /telescope requests.
✔️ Existing apps still use Gravatar when the new `config('telescope.avatar_driver')` key is not defined so this a non-breaking change for a point release.
✔️ Apps can now show local avatars taken from your own image hosting rather than using a remote service.

### Configuration

```php
namespace App\Providers;

// ...

class TelescopeServiceProvider extends TelescopeApplicationServiceProvider
{
    public function register()
    {
        // ...

        Telescope::avatar(function ($id) {
            $user = User::find($id);

            return url("/assets/images/users/{$user->avatar_path}");
        });
    }

    // ...
}
```
`config('telescope.avatar_driver')` options:

1. `'gravatar'` is the default
2. `'custom'` allows you to choose a URL path local to your application by registering a `Telescope::avatar()` callback.
3. Empty `''` (or any random string) will hide avatars in the UI and not call Gravatar's API.

### Remote Service Example

To switch from Gravatar to [Libravatar](https://wiki.libravatar.org/api/):

```php
Telescope::avatar(function ($id, $email) {
    $hash = md5(Str::lower($email));

    return vsprintf('https://seccdn.libravatar.org/avatar/%s?s=200&d=%s', [
        $hash,
        urlencode(url('/assets/image/avatar-placeholder.png')),
    ]);
});
```

### Implementation

The avatar URL is generated in PHP, only for JSON routes `GET /telescope/telescope-api/{type}/{telescopeEntryId}` for a preview of a single entry. i.e., this will not generate N+1 database queries when viewing a list of Telescope entries.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -82,6 +82,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Telescope Avatar Service
+    |--------------------------------------------------------------------------
+    |
+    | This option may be used to control how to show an avatar for authorized
+    | users. 'gravatar' will lookup an avatar by the user's email address.
+    | 'custom' requires registering a callback via Telescope::avatar().
+    |
+    */
+
+    'avatar_driver' => 'gravatar',
+
+    /*
+    |--------------------------------------------------------------------------
     | Ignored Paths & Commands
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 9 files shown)


taylorotwell:
Changed this so that just registering a callback with Telescope::avatar(fn) will enable custom driver. No need for configuration option.

PR #894, opened 2020-05-11
[3.x] Add afterStore hooks

barryvdh:
Add a hook for after the actual recording. This could be used for notifications etc, sending the actual Telescope URL in the mail/Slack message

```php
Telescope::afterStore(function(Telescope $telescope, $batchId) {

    foreach ($telescope::$entriesQueue as $entry) {
        if ($entry instanceof IncomingExceptionEntry) {
           dispatch(new TelescopeExceptionNotification($entry));
        }
    }
});
```

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -51,7 +51,7 @@ class Telescope
      * @var \Closure
      */
     public static $afterStoreHooks = [];
-    
+
     /**
      * The callbacks that add tags to the record.
      *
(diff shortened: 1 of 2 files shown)


taylorotwell:
Should we send the entries to this callback? It feels odd to have to dig into a static property of Telescope to access them.

barryvdh:
Yeah sure, so entries first and then batchId? Not sure if the instance is needed, but that was more inline with the afterRecord one. Will change it

taylorotwell:
Yeah I think that makes sense.

barryvdh:
I've updated the signature and also moved the store/filter methods inside the 'without recording' block, because otherwise it would generate Entries for the store/updates (which were then never flushed anyway).

PR #896, opened 2020-05-13
Clean up unused moontoast/math dependency

BenMorel:
AFAICS, `moontoast/math` is not used anymore in this project.
This is confirmed by all tests passing.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,6 @@
         "php": "^7.2",
         "ext-json": "*",
         "laravel/framework": "^6.0|^7.0",
-        "moontoast/math": "^1.1",
         "symfony/var-dumper": "^4.4|^5.0"
     },
     "require-dev": {


BenMorel:
Scratch that, this is an implicit dependency of `ramsey/uuid` < `4.0`.

jonnott:
> Scratch that, this is an implicit dependency of `ramsey/uuid` < `4.0`.

In which case this package should also have a specified dependency on `ramsey/uuid: ^3.0 | ^4.0`.

BenMorel:
It should actually be switched to `ramsey/uuid: ^4.0` to avoid this issue: https://github.com/ramsey/uuid/issues/288

jonnott:
> It should actually be switched to `ramsey/uuid: ^4.0` to avoid this issue: [ramsey/uuid#288](https://github.com/ramsey/uuid/issues/288)

But I guess that doesn't working because ramsey/uuid 4.x is only supported in Laravel 7.x, but they want Telescope 3.x to keeping supporting older Laravel versions (until Telescope 4.0 ships, at least)

driesvints:
This is all already fixed on the master branch for the upcoming 4.0 release. It's just a warning, ignore it for now.

PR #914, opened 2020-07-18
telescope:prune support for cockroachdb

kocoten1992:
Hi guy,

postgresql using ctid as an internal column since https://github.com/laravel/framework/pull/29393

some other DB also using postgres wire but do not explicit have ctid column (internally).

when rewrite sql this way, we can also support cockroachdb and this should not cause much perf impact.


taylorotwell:
We don't officially support Cockroach DB.

md5:
Another problem with using `ctid` shows up when the rows in `telescope_entries` are being constantly updated under heavy exception load, as mentioned in https://github.com/laravel/telescope/issues/1251

Since the `ctid` is being constantly changed by `UPDATE` queries in this condition, the prune never actually prunes anything for the `family_hash` values that are being heavily updated.

PR #915, opened 2020-07-21
Fixed frontend to work in subpath

edgardmessias:
This pull request solve the route problem in frontend when is served in a subpath.
Currently, the frontend try call from root + `config('telescope.path')`, but the correct need respect current base_path.

Example: 
URL base: `http://localhost/laravel/test`
The original request try call `http://localhost/telescope-api/....`
With this path: call `localhost/laravel/test/telescope-api/....`

### DIFF
--- a/resources/js/app.js
+++ b/resources/js/app.js
@@ -20,7 +20,7 @@ window.Popper = require('popper.js').default;
 
 moment.tz.setDefault(Telescope.timezone);
 
-window.Telescope.basePath = '/' + window.Telescope.path;
+window.Telescope.basePath = window.Telescope.path;
 
 let routerBasePath = window.Telescope.basePath + '/';
 
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -729,7 +729,7 @@ class Telescope
     public static function scriptVariables()
     {
         return [
-            'path' => config('telescope.path'),
+            'path' => parse_url(route('telescope'), PHP_URL_PATH) ?? '/',
             'timezone' => config('app.timezone'),
             'recording' => ! cache('telescope:pause-recording'),
         ];


driesvints:
We don't support using Laravel in a subdirectory. Please don't place your Laravel installation in a subdirectory as it exposes your app to the outside.

edgardmessias:
@driesvints, I think you got me wrong.

I not exposes my laravel, I have used apache config and reverse proxy to use laravel in a subpath, like this apache config:

```
# short example
Alias "/test/" "/<secure folder>/<laravel project>/public/"
```

driesvints:
Why can't you set `telescope.path` to `laravel/test/telescope-api` in this case?

edgardmessias:
Because of route prefix in TelescopeServiceProvider.php: https://github.com/laravel/telescope/blob/53c40e89f0de45d878665f17f2da02a05c18a394/src/TelescopeServiceProvider.php#L61

If I set `telescope.path` to `laravel/test/telescope-api`, the route of frontend will change to `laravel/test/telescope-api`, and the API will try call: `/laravel/test/telescope-api/telescope-api`, but because of "proxy or alias", the correct final route is `laravel/test/laravel/test/telescope-api`.

edgardmessias:
It is hard to understand, but when you set `telescope.path`, you are changing the prefix of routing, but, the `scriptVariables` return a absolute URL instead of a resolved route.

If you look [src/Illuminate/Routing/RouteUrlGenerator.php#L102](https://github.com/laravel/framework/blob/0d8af41753a3ed5a6543e9547dc8b68098ac621e/src/Illuminate/Routing/RouteUrlGenerator.php#L102) there is baseUrl when you use "route function"

driesvints:
@edgardmessias we're gonna let this one be. Sorry.

luanmm:
@driesvints there is no argument possible to keep insisting that this change **exposes** the application (in a security point-of-view) in all its uses. Or the only supported use-case for Laravel is a simple website? This kind of thought is just incompatible with the size of Laravel framework. Why should you impose limitations on this, when the web is made of a diversity of architectures?

I'm using Laravel in a system that uses lot of projects as micro-front/backends. So, in my case, its just **impossible** to expose all the Laravel apps in the root URL of the system/application. Every Laravel application has its own path inside the domain. And this doesn't even relate with any security concerns, because every Laravel instance runs its own "root-level" application. **Nothing** is exposed in the sense of security, what seems to be your concerns here. But it **must** be "exposed", by the server (reverse proxy), as a sub-path.

This PR is just allowing more advanced architectures to work with Telescope. Without it, these projects that aren't so trivial just can't use Telescope (without workarounds at least).

Issue #919, opened 2020-07-24
Query detail page shows mysql variables as "undefined"


semiherdogan:
- Telescope Version: 2.1.7
- Laravel Version: 6.18.22
- PHP Version:  7.3.14
- Database Driver & Version:
mysql & 5.7.28
### Description:
When i use mysql variable in sql query telescope query detail page shows that variable as undefined.

### Steps To Reproduce:
Run this query and check telescope query detail page.
`\DB::select('select @user_id := id as id from users');`

It shows correctly at list page
![image](https://user-images.githubusercontent.com/5646041/88387073-64625b00-cdba-11ea-8ed5-da7cf6fcb37e.png)

But in detail shows like this:
![image](https://user-images.githubusercontent.com/5646041/88387114-77752b00-cdba-11ea-8a7a-2c98f546fec8.png)

driesvints:
Seems like Telescope has some problems with session variables. Appreciating help with this or a PR until we get to this ourselves.

derekmd:
The Telescope Vue template uses NPM package [sql-formatter](https://www.npmjs.com/package/sql-formatter) to highlight SQL queries. Digging into its code, the `@` character is treated similar to the `?` placeholder, but named. MS SQL Server, etc. support this syntax.

https://github.com/zeroturnaround/sql-formatter/blob/1062bbc278319d6bf0e2d4cc30fba427411a6dab/src/languages/StandardSqlFormatter.js#L100

With a variable _assignment_ expression, the library is trying to pull a binding value that doesn't exist.

```javascript
sqlFormatter.format('select @user_id := id as id from users', {params: []});
// select
//   undefined: = id as id
// from
//   users
```

It expects a named options parameter, e.g., `{params: {user_id: 1}`

Excluding the `params` option will stop the library from attempting to fill the placeholder.

```javascript
sqlFormatter.format('select @user_id := id as id from users');
// select
//   @user_id: = id as id
// from
//   users
```

So this branch could be submitted to fix the above case: https://github.com/laravel/telescope/compare/3.x...derekmd:fix-mysql-variable-dump

### NPM package `sql-formatter` needs a MySQL fix

[...]

PR #923, opened 2020-07-27
[3.x] Checked query parameters before send to sqlFormatter. fixes #919

semiherdogan:
**This pull request fixes sql formatter that shows mysql variables as undefined.**

For this query `\DB::select('select @user_id := id as id from users');` telescope query page shows like this:

image

With this pr you can see mysql variables like image below:
image

For fixing this issue i have checked if params exists before send to sqlFormatter.

You may see related conversation here. https://github.com/laravel/telescope/issues/919

Fixes #919

PR #924, opened 2020-07-27
[3.x] Remove problematic unused code when formatting SQL

derekmd:
Fixes: https://github.com/laravel/telescope/issues/919
Replaces pull request https://github.com/laravel/telescope/pull/923 as per [Dries' request](https://github.com/laravel/telescope/pull/923#discussion_r461064565)

Formatted SQL preview:

before

which should be:

after

Telescope's `QueryWatcher` always stores an empty array for query parameter bindings.

https://github.com/laravel/telescope/blob/f5dee9aa18f4ec426130676bd1e50dac79ec91e4/src/Watchers/QueryWatcher.php#L39-L41

This causes an issue in NPM package `sql-formatter` when the query has MySQL session variables referenced with an `'@'` character. Using the generic SQL:2011 dialect, it's treated like a named placeholder (e.g., for MS SQL Server.)

```javascript
sqlFormatter.format('select @user_id := id as id from users', {params: []});
// select
//   undefined: = id as id
// from
//   users
```

This expects a named options parameter as if `@user_id` is referenced within a `WHERE` clause, e.g., `{params: {user_id: 1}}` Missing a named binding value, JavaScript `undefined` is instead filled in.

Excluding the `params` option will stop the package from attempting to fill the placeholder.

```javascript
sqlFormatter.format('select @user_id := id as id from users');
// select
//   @user_id: = id as id
// from
//   users
```

### `<query-preview>` cleanup

1. stop passing the bindings empty array to the `.format()` method since the argument is _never_ used.
   * `slotProps.entry.content.bindings` returned by the API endpoint is always `[]`
   * should that QueryWatcher.php `'bindings' => []` line also be removed? Does the JSON payload need that when PDO parameters are replaced before database insertion?
     * The change was from https://github.com/laravel/telescope/pull/706/files#diff-fa8840834f665959fe3cfa98a78adadcR42 so I think that line has been redundant since Telescope v2.1 (Sept 3, 2019)
2. Instead of referencing the generic wrapper `sqlFormatter.format()` [public API](https://github.com/zeroturnaround/sql-formatter/blob/master/src/sqlFormatter.js#L29), pull in the exact `StandardSqlFormatter` implementation. ES6 classes `Db2Formatter`, `N1qlFormatter`, & `PlSqlFormatter` are also being unnecessarily bundled into Telescope's app.js. This shrinks the production file 813kb to 796kb. There's definitely more room for perf improvement here. e.g.,
   * remove `/**!` NPM package comments
   * remove bootstrap.js components Telescope doesn't use like Carousel
3. remove unused Vue component props `entry` & `batch` since query data is accessed via a slotted scope.

### DIFF
--- a/resources/js/screens/queries/preview.vue
+++ b/resources/js/screens/queries/preview.vue
@@ -1,24 +1,16 @@
 <script type="text/ecmascript-6">
-    import _ from 'lodash';
-    import sqlFormatter from "sql-formatter";
     import hljs from 'highlight.js/lib/highlight';
     import sql from 'highlight.js/lib/languages/sql';
+    import StandardSqlFormatter from 'sql-formatter/src/languages/StandardSqlFormatter';
 
     export default {
         data(){
             return {
-                entry: null,
-                batch: [],
+                sqlFormatter: new StandardSqlFormatter({}),
             };
         },
 
         methods:{
-            formatSQL(sql, params) {
-                return sqlFormatter.format(sql, {
-                    params: _.map(params, param => _.isString(param) ? '"'+param+'"' : param)
-                });
-            },
-
             highlightSQL() {
                 this.$nextTick(() => {
                     hljs.registerLanguage('sql', sql);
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @derekmd

PR #949, opened 2020-09-17
Retain exceptions data

sent316:
Takes —keep-exceptions argument


driesvints:
Thanks for your pr. There's a few breaking changes here so please send to master. Also, there's quite some code style violations. We don't usually don't use scalars as type hints and the docblocks don't conform to the rest of the code base. Please make sure you update these.

PR #950, opened 2020-09-17
Retain exceptions data

sent316:
Takes —keep-exceptions argument


src/Storage/DatabaseEntriesRepository.php
```diff
@@ -335,11 +335,15 @@ public function stopMonitoring(array $tags)
      * @param  \DateTimeInterface  $before
      * @return int
      */
-    public function prune(DateTimeInterface $before)
+    public function prune(DateTimeInterface $before, $keepExceptions)
```
driesvints, line 338: Update the docblock for this param

sent316: Updated doc blocks for the parameter

Issue #955, opened 2020-09-29
500 Error with messsage Type of Laravel\\Telescope\\Storage\\EntryModel::$table must be string


puncoz:
- Telescope Version: v4.0.1
- Laravel Version: v8.6.0
- PHP Version: 7.4.10
- Database Driver & Version: 

### Description:
A 500 Internal Server Error thrown while calling api `/telescope/telescope-api/requests` by telescope dashboard.

Error Response:

```
{
    "message": "Type of Laravel\\Telescope\\Storage\\EntryModel::$table must be string (as in class Illuminate\\Database\\Eloquent\\Model)",
    "exception": "Symfony\\Component\\ErrorHandler\\Error\\FatalError",
    "file": "/path/to/application/vendor/laravel/telescope/src/Storage/EntryModel.php",
    "line": 182,
    "trace": []
}
```

### Steps To Reproduce:
- Setup the telescope
- Browse telescope dashboard `/telescope` in browser, it will keep on loading while we can see the 500 Internal Server Error in console.

paras-malhotra:
That's something wrong in your setup, nothing wrong with Telescope per se.

driesvints:
We don't use typed properties.

PR #956, opened 2020-09-29
issue of type defination in laravel 8 with php7.4

puncoz:
Issue related #955 fixed by defining proper string type declaration of Model's table property.


driesvints:
We don't use type hints on properties.

Issue #636, opened 2019-05-22
[Feature request] Collapse dumps


AndreasHerss:
It would be nice to be able to collapse the dumps made, as i often dump collections, but as it is right now, i cant collapse data, it's all open.

I would like for it to be like when using dd() and chrome's renderings where i can collapse all childs.

This makes it much easier to inspect data, when not all relations are expanded :)

taylorotwell:
This is not a bug report so I am closing for now.

PR #959, opened 2020-10-05
[4.x] Dump visibility toggling, search, etc.

jasonvarga:
This PR hooks up the var dumper JS so that the toggles and search work.

Implements #636 

The most valuable part of this IMO is that if you dump an object, your page isn't full of completely expanded objects. (Try dumping `app()` a couple of times 😅 ) It's more consistent with what you'd see if you weren't using Telescope.

hNUVQ0

Since Vue can't execute JS that's sent from the server and injected through `v-html`, I've had to basically copy paste [the var dumper JS](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L163-L656).

There's only a handful of modifications:

- Instead of `Sfdump = window.Sfdump || (function (doc) {`, I'm putting the function into a `this.sfDump` property on the Vue instance.
- The quotes [here](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L529-L538) have been replaced by JS template strings so it could compile.
- Added a `let` before `SearchState` [here](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L470) to get around a `ReferenceError: assignment to undeclared variable SearchState` error.
- Added `let node;` above [this loop](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L575) to get around a `assignment to undeclared variable node` error.
- Instead of string replacing dynamic `$options` [here](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L304), I've just hardcoded in [the defaults](https://github.com/symfony/symfony/blob/5.1/src/Symfony/Component/VarDumper/Dumper/HtmlDumper.php#L70-L74).

### DIFF
--- a/resources/js/screens/dumps/index.vue
+++ b/resources/js/screens/dumps/index.vue
@@ -1,5 +1,6 @@
 <script type="text/ecmascript-6">
     import axios from 'axios';
+    import sfdump from './sfdump';
 
     export default {
         /**
@@ -12,7 +13,9 @@
                 ready: false,
                 newEntriesTimeout: null,
                 newEntriesTimer: 2000,
-                recordingStatus: 'enabled'
+                recordingStatus: 'enabled',
+                sfDump: null,
+                triggered: [],
             };
         },
 
@@ -22,6 +25,8 @@
          */
         mounted() {
             document.title = "Dumps - Telescope";
+
+            this.initDumperJs();
 
             this.loadEntries();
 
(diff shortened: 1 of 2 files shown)


jasonvarga:
It would be nice if StyleCI could ignore that sfdump file so we can keep it as close to the original as possible. I'll try to figure that out.

jasonvarga:
Or not. Thanks!

driesvints:
@jasonvarga think it's best that we keep all code style consistent.

PR #964, opened 2020-10-07
3.x Feature: Be able to specify only allowed paths

u12206050:
Non-breaking changes. (For version 3.x)

Only start recording on the given paths if any are specified
Reason: It is a hassel if you only want to monitor a single path.

Example: Great for only monitoring api requests in production eg.:

    'only_allow_paths' => [
        'api/*'
    ],
If none are specified then it falls back to checking ignore_paths which is still the default.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -82,6 +82,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Allowed Paths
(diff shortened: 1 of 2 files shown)


taylorotwell:
Telescope 4 is the currently maintained branch so please send this there if you are interested.

PR #971, opened 2020-10-19
[4.x] Added isCommand() filter

Ali-Shaikh:
Determine if the incoming entry is an artisan command.

### DIFF
--- a/src/IncomingEntry.php
+++ b/src/IncomingEntry.php
@@ -182,6 +182,16 @@ class IncomingEntry
     }
 
     /**
+     * Determine if the incoming entry is an artisan command.
+     *
+     * @return bool
+     */
+    public function isCommand()
+    {
+        return $this->type === EntryType::COMMAND;
+    }
+
+    /**
      * Determine if the incoming entry is a request.
      *
      * @return bool


taylorotwell:
The type property is public. You can already inspect it.

Ali-Shaikh:
> The type property is public. You can already inspect it.

Hi Taylor, thanks for reviewing the PR. I was building upon the existing filters added in the 'src/IncomingEntry.php' file and added these as they were missing. These filters allow builtin filtering capabilities when using them in the TelescopeServiceProvider. I needed to filter some entries in my application and that's why I added these so others could use it too. 

I am just curious as to why the PRs were closed instead of being merged. Thanks

paras-malhotra:
I think what he meant to say was that you can use `$entry->type === EntryType::COMMAND` directly in your `TelescopeServiceProvider` since `type` is a public property.

Ali-Shaikh:
> I think what he meant to say was that you can use `$entry->type === EntryType::COMMAND` directly in your `TelescopeServiceProvider` since `type` is a public property.

Got it, thanks. I just thought that using isCommand() way is much cleaner and consistent which is why I created the PRs

Issue #978, opened 2020-11-04
Telescope doesn't publish migrations in fresh Laravel install


mrjutsu:
- Telescope Version: 4.3.0
- Laravel Version: 8.13.0
- PHP Version: 7.4.1
- Database Driver & Version: MySQL 5.7.26

### Description:
Running `php artisan vendor:publish --tag=telescope-migrations` after installing Telescope in a fresh Laravel project throws the error: `Can't locate path: <.../vendor/laravel/telescope/src/Storage/migrations>`

### Steps To Reproduce:
Create a new Laravel project:

`laravel new project`

Run `composer require laravel/telescope`

Add line `TELESCOPE_ENABLED=true` to .env file.

Run `php artisan telescope:install`. The files `TelescopeServiceProvider.php` and `telescope.php` are published in the Providers and config directories respectively but the project's migrations directory doesn't show the Telescope migrations.

Running `php artisan vendor:publish --tag=telescope-migrations` throws the aforementioned error.

I also noted that manually changing line 87 at `src/TelescopeServiceProvider.php` in Telescope's source code from

`__DIR__.'/Storage/migrations' => database_path('migrations'),`

to

`__DIR__.'/../database/migrations' => database_path('migrations'),`

fixes the issue and correctly publishes Telescope's migrations.

timhaak:
I hit the same problem.

Created a pull request to hopefully save someone some time :) 

laravel/telescope#979

driesvints:
Thanks. Merged and released.

PR #979, opened 2020-11-05
[4.x] Fix for Can't locate path: <.../vendor/laravel/telescope/src/Storage/migrations>

timhaak:
Fix for Can't locate path: <.../vendor/laravel/telescope/src/Storage/migrations>

Hit the same problem as laravel/telescope#978

I've manually fixed my side but thought it might help to do a pull request to hopefully save you guys some work.


driesvints:
Thanks! I'll release this in a bit.

PR #985, opened 2020-11-23
[4.x] Fixed file not found error during install command

myckhel:

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -74,7 +74,7 @@ class InstallCommand extends Command
         file_put_contents(app_path('Providers/TelescopeServiceProvider.php'), str_replace(
             "namespace App\Providers;",
             "namespace {$namespace}\Providers;",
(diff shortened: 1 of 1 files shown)


driesvints:
This is incorrect. You're referencing the TelescopeProvider from the package, not the one intended for the app. 

The app one is published: https://github.com/laravel/telescope/blob/4.x/src/TelescopeServiceProvider.php#L98-L100
Through here: https://github.com/laravel/telescope/blob/4.x/src/Console/InstallCommand.php#L32
And after that namespaces are optionally renamed here: https://github.com/laravel/telescope/blob/4.x/src/Console/InstallCommand.php#L74-L78

myckhel:
Somehow the intended `Providers/TelescopeServiceProvider` was not found
When it got to this line https://github.com/laravel/telescope/blob/7e9d9fdab18b86db6c8368f2a026ff695c8287e9/src/Console/InstallCommand.php#L77

Ran the install command multiple times but kept getting same error
Untill i made the pushed changes before it worked.

Issue #534, opened 2019-01-17
BUG: Database migration creating table without a primary key


tomswinkels:
I have created the database with `php artisan migrate` and not all the tables has some primary key.

`SQLSTATE[HY000]: General error: 1105 Percona-XtraDB-Cluster prohibits use of DML command on a table (prepr_api.telescope_entries_tags) without an explicit primary key with pxc_strict_mode = ENFORCING or MASTER (SQL: insert into `telescope_entries_tags` (`entry_uuid`, `tag`) values (8cc397bb-ea30-42b3-8705-00f6eefb169d, failed))`

themsaid:
What is the database engine?

rs-sliske:
i believe percona is a mysql fork

tomswinkels:
Yeah we use percona.

tomswinkels:
@ybaruchel why do you have close that branch?

ybaruchel:
@tomswinkels Because I accidentally added a primary key to the wrong table and it broke the build.. I didn’t had time to fix it so I closed the PR

Elycin:
Bump to this.

My workplace is interested in using telescope for diagnostic reasons and we're having this current issue with our percona cluster. Would love to see this implemented.

PR #986, opened 2020-11-24
Add primary keys to database tables

burgercode:
It is considered good practice to always set a primary key for improved performance. As such, some managed databases, like [DigitalOcean, requires the definition of a primary key](https://www.digitalocean.com/docs/databases/mysql/how-to/create-primary-keys/) on each table in every new MySQL database. However, Telescope doesn't have any primary key for the following tables:

- `telescope_entries_tags`
- `telescope_monitoring`

Thus using Telescope in projects using a managed database like DigitalOcean is impossible.

This pull request adds the missing primary keys without affecting the tables' structure (i.e., no additional columns are added).

Fixing #534 Issue

### DIFF
--- a/database/migrations/2018_08_08_100000_create_telescope_entries_table.php
+++ b/database/migrations/2018_08_08_100000_create_telescope_entries_table.php
@@ -61,6 +61,7 @@ class CreateTelescopeEntriesTable extends Migration
             $table->uuid('entry_uuid');
             $table->string('tag');
 
+            $table->primary(['entry_uuid', 'tag']);
             $table->index(['entry_uuid', 'tag']);
             $table->index('tag');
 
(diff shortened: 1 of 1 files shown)


burgercode:
Apparently, this won't fix the bug with DigitalOcean as we can see here: https://github.com/laravel/framework/issues/33238

Laravel first creates the table and then sets the key as primary. However, I think this PR is still relevant since it is good practice in general.

taylorotwell:
If you're adding primary keys I think you need to remove those `index` calls. Looks like duplicate indexes are being added.

Also, as you said, this won't fix the DigitalOcean problem as that is more of a limitation on their end.

FrittenKeeZ:
Would it not be easier to simply add `$table->id();` to those two tables?

PR #991, opened 2020-12-07
[4.x] Updated db command for clearing tables

Theograms:
Hello, while clearing the db tables the `delete` statement being used is significantly slower than `truncate`. 

The logs the `delete` keeps are useless since these tables contain only debugging info and not essential data.

While using `telescope:clear` it takes very much time for several million records whilst manually executing `truncate` takes a couple of seconds for about the same amount of records. Also it consumes fewer server resources. I think the statement should be changed from `delete` to `truncate`.

[Truncate Statement](https://dev.mysql.com/doc/refman/8.0/en/truncate-table.html)
[Truncate vs Delete](https://stackoverflow.com/questions/20559893/comparison-of-truncate-vs-delete-in-mysql-sqlserver)

### DIFF
--- a/src/Storage/DatabaseEntriesRepository.php
+++ b/src/Storage/DatabaseEntriesRepository.php
@@ -358,8 +358,8 @@ class DatabaseEntriesRepository implements Contract, ClearableRepository, Prunab
      */
     public function clear()
     {
-        $this->table('telescope_entries')->delete();
-        $this->table('telescope_monitoring')->delete();
+        $this->table('telescope_entries')->truncate();
+        $this->table('telescope_monitoring')->truncate();
     }
 
     /**


Namoshek:
@Theograms @taylorotwell 
This breaks the command when using an SQL Server database:
> Msg 4712, Level 16, State 1, Line 1
> Cannot truncate table 'dbo.telescope_entries' because it is being referenced by a FOREIGN KEY constraint.

I do not have much experience with the other supported DBMS, so I cannot tell if similar issues arise during their usage as well.

driesvints:
I'm reverting this

louisgab:
@driesvints @taylorotwell May I suggest:

```php
DB::connection($this->connection)->enableForeignKeyConstraints();
$this->table('telescope_entries')->truncate();
$this->table('telescope_monitoring')->truncate();
DB::connection($this->connection)->disableForeignKeyConstraints();
```
Maybe in a separate command `telescope:truncate` so it can be carefully used on supported DBMS ?

I would happily PR it

driesvints:
We're not reconsidering this atm sorry.

PR #995, opened 2020-12-09
[4.x] Support JSON queries on telescope entries

paras-malhotra:
Currently, Telescope uses the longText column type for content. However, databases such as PostgreSQL do not support JSON queries (with the `->` notation in Laravel) unless the data type is JSON. A simple use case is grouping the content by line number and file for slow queries, to see where the slowest queries appear in the code.

Moreover, even for databases like mysql, the native JSON type is more efficient for json queries, even though they are also supported on the text data type.


taylorotwell:
Telescope includes support for disabling the default migrations so you can run your own: https://laravel.com/docs/8.x/telescope#migration-customization

So you could make this change in your own application.

PR #999, opened 2020-12-12
[4.x] Add Model Hydration Watcher

paras-malhotra:
### Motivation
While debugging, it is very useful to know how many models are being hydrated for a request - something that can have a significant on performance / memory usage. So, I've added a model hydration watcher that records the number of models hydrated for each model. Inspired from a similar popular feature of Debugbar: https://github.com/barryvdh/laravel-debugbar/pull/947

### Index Screen
hydrations

### Preview Screen
hydrations-preview

### Related Entries
hydrations-related-entries

### Note
This PR would need a re-build and I haven't committed the compiled files (just the source) as per the contribution guidelines.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -1,6 +1,7 @@
 <?php
 
 use Laravel\Telescope\Http\Middleware\Authorize;
+use Laravel\Telescope\Storage\EntryModel;
 use Laravel\Telescope\Watchers;
 
 return [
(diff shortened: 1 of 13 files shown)


driesvints:
The separate menu screen feels a bit out of place. Any chance we can move this under the "Models" menu item somehow?

paras-malhotra:
@driesvints the model watcher is structured to record every event, whereas the hydration watcher aggregates the events (sums the count). So, the logic and also fields for both are quite different that's why I had it as a separate watcher.

taylorotwell:
I agree I don't want to add an entire tab for this.

PR #1005, opened 2021-01-22
Prevent access to uninitialized typed properties

### LINKED ISSUE #1002: Error parsing Job class with uninitialized typed properties
- Telescope Version: 4.4.0
- Laravel Version: 8.19.0
- PHP Version: 7.4.13

### Description:
When a Job class contains typed properties that are not initialized with a default value nor set in the constructor, an error is thrown: "Typed property x must not be accessed before initialization".

I believe to solve this in a backward compatible way, it would require something like this in `Laravel\Telescope\ExtractProperties::from`:
```php
if (version_compare(phpversion(), '7.4', '>=') && !$property->isInitialized($target)) {
   return [$property->getName() => null];
} else {
	...
```
Not sure if you would rather return some type of `UninitializedProperty` object instead so that this could actually be distinguished from `null` in the UI.

### Steps To Reproduce:
```php
    public function test_extract_properties_from_class_with_uninitialized_typed_properties()
    {
        $properties = \Laravel\Telescope\ExtractProperties::from(
            new class {
                private string $uninitialized;
            },
        );

        $this->assertEquals(['uninitialized' => null], $properties);
    }
```

wfeller:
Fixes #1002

### DIFF
--- a/src/ExtractProperties.php
+++ b/src/ExtractProperties.php
@@ -20,6 +20,10 @@ class ExtractProperties
         return collect((new ReflectionClass($target))->getProperties())
             ->mapWithKeys(function ($property) use ($target) {
                 $property->setAccessible(true);
+                
+                if (PHP_VERSION_ID >= 70400 && ! $property->isInitialized($target)) {
+                    return [];
+                }
 
                 if (($value = $property->getValue($target)) instanceof Model) {
                     return [$property->getName() => FormatModel::given($value)];


driesvints:
Again: this isn't a Telescope issue. This is PHP telling you you need to initialize your object properties.

taylorotwell:
I do indeed think this is a Telescope issue and we have handled this issue in other places before.

taylorotwell:
Telescope need not require you to initialize every property on a job just to use Telescope.

PR #1009, opened 2021-02-02
Add MongoDB compatibility

saldomik:
In Zanichelli we use Telescope to monitor several web applications.

When we realized the lack of out-of-the-box compatibility with MongoDB, we thought it would be a good idea to propose our way to use Telescope in a MongoDB based application.

This solution doesn't affect any previous functionality or tests and allows the user to choose either MYSQL or MongoDB.

We used Jenssegers package to provide a compatibility layer for MongoDB, creating two classes MongoModel and RegularModel; the first one extends Jenssegers Model, the second one extends Eloquent Model.
We also changed the EntryModel class in order to be a sort of entry point, now it only contains a static method that returns the correct class based on the database version specified in the env file.

In case one wanted to use Telescope with MongoDB, it is only necessary to run `composer require jenssegers/mongodb` e set 'mongodb' as DB_CONNECTION inside project env file.


taylorotwell:
We haven no plans take on the Mongo maintenance burden internally.

PR #1034, opened 2021-03-16
[4.x] Move booting of commands

driesvints:


src/TelescopeServiceProvider.php
```diff
@@ -27,6 +27,7 @@ public function boot()
         $this->registerRoutes();
         $this->registerMigrations();
         $this->registerPublishing();
+        $this->registerCommands();
```
Namoshek, line 30: @taylorotwell @driesvints This change can actually break (Docker) builds and deployments if Telescope is not enabled (line 21), because the command is not registered anymore.

driesvints: Thanks, I sent in a PR for this: https://github.com/laravel/telescope/pull/1045

PR #1039, opened 2021-03-22
[4.x] Update to Mix v6

driesvints:
This updates our build pipeline to Laravel Mix v6. Also bumps several other dependencies. I'll send in a separate PR later on to update Vue to v3.

I needed to remove the `copy` call from the Laravel Mix file because it was failing for some reason. I also find it a bit out that a directory is created outside this repo. I guess this was for testing purposes? 

Closes https://github.com/laravel/telescope/issues/1019


tomirons:
@driesvints with the update to `sql-formatter` there are now whitespaces where they shouldn't be. You can fix this by setting the `language` on the formatter to `mysql` but I'm not sure how it would act with a _postgres_ database.

Here is the new output:
```sql
select
  *
from
  ` database `.` table `
where
  ` database `.` table `.` column ` = 1
limit
  1
```

PR #1040, opened 2021-04-02
Prep For Octane

taylorotwell:
This allows Telescope to work with Laravel Octane.


caiokawasaki:
@taylorotwell I think the answer will be no, but is there any way to log things from octane/swoole workers in octane?

driesvints:
@caiokawasaki not right now

PR #1045, opened 2021-04-07
[4.x] Always boot commands

driesvints:
In some cases you'd always want to boot commands, even if Telescope itself isn't enabled.


Namoshek:
Thank you!

Issue #1027, opened 2021-03-10
Filter requests per more than one tag


LeonardoLpds:
Should be nice if we can filter requests per more than one tag.
For instance, I created a tag for uri, so I can filter **uri:/login** then I created a tag for client id to filter requests like that **id:123**, What I want to do is search both tags simultaneously

driesvints:
Heya, if you want to propose something like this, just feel free to send in a PR so we can look at the code needed to make this happen. Thanks

PR #1055, opened 2021-05-01
[4.x] - Filter many tags

Lorrhanl:
In reference the issue #1027, I have a simple idea.

in search tag, separate for comma for search more than one.

it helps in the search for many requests with specific tags.


taylorotwell:

PR #1060, opened 2021-05-10
"operator does not exist: boolean = integer" error for should_display_on_index field

omerkoseoglu:
(PDOException(code: 42883): SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: boolean = integer


omerkoseoglu:
When using PostgreSQL, although it is boolean in migration, it sends a numeric value as a query.

PR #1063, opened 2021-05-19
Remove asset manifest check and warning

### LINKED ISSUE #1062: 'Assets not published' exception thrown with default Vapor setup
- Telescope Version: 4.4.9
- Laravel Version: 8.41.0
- PHP Version: 8.0.3
- Database Driver & Version: n/a

### Description:

By default with a fresh Laravel app running on Vapor, it's not possible to view Telescope at all. The "Telescope assets are not published" exception is always thrown, because it checks `public_path('vendor/telescope/mix-manifest.json')` which doesn't exist on the lambda runner. Telescope **can** load its assets normally from S3/Cloudfront where Vapor uploads them, but it's impossible to get to that point because of the exception being thrown.

It is possible to work around this by adding `php artisan telescope:publish` to Vapor's build hooks and including all of Telescope's assets in the build artifact, but this adds about 1.5MB to the build and is redundant since the assets are _also_ always uploaded to S3/Cloudfront.

### Steps To Reproduce:

New Laravel app, install Telescope, deploy to Vapor, see the "Telescope assets are not published" exception if you try to view Telescope on Vapor.

bakerkretzmar:
This PR reverts #729, removing the checks to determine at runtime if Telescope's assets are published and up to date.

The exception that is thrown if the assets appear to be missing doesn't play nicely with Vapor because it assumes that the assets should be present on the local filesystem at runtime, and the warning feels unnecessary too since for quite a long time now the docs have had [very clear instructions about how to keep assets up to date automatically](https://laravel.com/docs/8.x/telescope#upgrading-telescope).

Closes #1062.

See also laravel/telescope#594 and laravel/horizon#554, this warning wasn't immediately popular when it was added and both Telescope and Horizon now provide documentation that make it unnecessary.

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -201,12 +201,6 @@
             </div>
 
             <div class="col-10">
-                @if (! $assetsAreCurrent)
-                    <div class="alert alert-warning">
-                        The published Telescope assets are not up-to-date with the installed version. To update, run:<br/><code>php artisan telescope:publish</code>
-                    </div>
-                @endif
-
                 <router-view></router-view>
             </div>
         </div>
--- a/src/Http/Controllers/HomeController.php
+++ b/src/Http/Controllers/HomeController.php
@@ -17,7 +17,6 @@ class HomeController extends Controller
         return view('telescope::layout', [
             'cssFile' => Telescope::$useDarkTheme ? 'app-dark.css' : 'app.css',
             'telescopeScriptVariables' => Telescope::scriptVariables(),
-            'assetsAreCurrent' => Telescope::assetsAreCurrent(),
         ]);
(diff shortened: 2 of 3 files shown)


barryvdh:
Can't we just remove the exception if no file is found, instead of not checking at all? If the version is mismatch it's easy to fix so don't really see what the problem is (outside of the exception).

driesvints:
@barryvdh Probably best that you propose something with a PR

bakerkretzmar:
Vapor UI does this in a really cool way actually, with a middleware: https://github.com/laravel/vapor-ui/blob/master/config/vapor-ui.php. That strikes me as the best of both worlds, basically makes it completely configurable by the user.

Issue #1065, opened 2021-05-27
Telescope breaks when attempt quote the bindings from an PDO_ODBC connection


juanparati:
- Telescope Version: 4.4
- Laravel Version: ^8.0
- PHP Version: 8.0
- Database Driver & Version: PDO_ODBC with unixODBC

### Description:
The following error is raised by Telescope when a query is recorded:

`PDOException with message 'SQLSTATE[IM001]: Driver does not support this function: driver does not support quoting'`

The reason is that telescope uses `$event->connection->getPdo()->quote($binding)` in order to binding the quotes, however according to [PHP documentation](https://www.php.net/manual/en/pdo.quote.php) PDO::quote is not supported by PDO_ODBC.

### Steps To Reproduce:

1. Setup a ODBC source in Laravel (You can install the package a [Laravel ODBC package](https://github.com/yoramdelangen/laravel-pdo-odbc)).
2. Activate Telescope.
3. Run a query against the ODBC connection.

### Notes

A possible solution to this problem it could be verify if PDO driver is different than ODBC, and use an alternative method of quoting the bindings.

It's possible to check the driver using:

```
DB::connection('odbc')
  ->getPdo()
  ->getAttribute(PDO::ATTR_DRIVER_NAME);
```

driesvints:
PDO_ODBC is not supported by Laravel, sorry.

driesvints:
Sorry, I'm mistaken. It is supported: https://laravel.com/docs/8.x/database#mssql-configuration

driesvints:
Can you send in a PR for that?

juanparati:
I will do it.

driesvints:
Thanks! :)

PR #1066, opened 2021-05-27
[4.x] Implemented fallback for PDO::quote method on QueryWatcher

juanparati:
This pull request will solve https://github.com/laravel/telescope/issues/1065.

Note: The fallback method may not constraint with all the target databases that ODBC is connected, It means that there is no way to know which charset, and quote symbol is the default used by the database, however the fallback method should represent an accurate query string for the most popular databases like mSQL, BigQuery or Snowflake.

### DIFF
--- a/src/Watchers/QueryWatcher.php
+++ b/src/Watchers/QueryWatcher.php
@@ -10,6 +10,7 @@ class QueryWatcher extends Watcher
 {
     use FetchesStackTrace;
 
+
     /**
      * Register the watcher.
      *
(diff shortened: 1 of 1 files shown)


src/Watchers/QueryWatcher.php
```diff
@@ -100,12 +100,39 @@ public function replaceBindings($event)
...
+     *
+     * @param  \Illuminate\Database\Events\QueryExecuted  $event
+     * @param string $binding
+     * @return string
+     */
+    protected function quoteStringBinding($event, $binding)
+    {
+        try {
+            return $event->connection->getPdo()->quote($binding);
+        } catch (\PDOException $e) {
```
driesvints, line 123: Import this object type.

juanparati:
@driesvints : Sorry driesvints, I did a last change.

juanparati:
😅 I have that last commit is accepted.

I replaced `$event->quoteStringBinding` by `$this->quoteStringBinding`

In other way the release is buggy.

driesvints:
@juanparati can you please do the changes I asked for above? Thanks

juanparati:
Done.

PR #1073, opened 2021-06-06
Add http client request watcher

gdebrauwer:
In Laravel 8.45.0 (https://github.com/laravel/framework/pull/37572), events were introduced in the Http Client. I thought it would be nice to add a watcher for the Http Client, using these new events. This will make it easier to debug requests/responses of the Http Client.

In the detail view, you can see the request payload and response body, but also the headers of both the request and the response.

The request payload and headers are being filtered based on the `Telescope::$hiddenRequestParameters`, similar to a regular Laravel request in Telescope.

The request body is converted to json (just as as is the case with a regular Laravel request in Telescope). There is one limitation currently in case of a multipart request. If the request contains a file without a filename or headers, then that file will not be converted to a json object containing name and size and headers. Instead the full file content will still be included in the saved request payload. This is because I don't think there is a way to detect if "contents" in multipart item is a file or not.

Schermafbeelding 2021-06-08 om 20 19 11
Schermafbeelding 2021-06-08 om 20 19 29


resources/js/components/RelatedEntries.vue
```diff
@@ -579,6 +586,49 @@
...
+                    </td>
+
+                    <td class="table-fit" :data-timeago="entry.created_at" :title="entry.created_at">
+                        {{timeAgo(entry.created_at)}}
+                    </td>
+
+                    <td class="table-fit">
+                        <router-link :to="{name:'client-request-preview', params:{id: entry.id}}" class="control-action">
+                            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 22 16">
+                                <path d="M16.56 13.66a8 8 0 0 1-11.32 0L.3 8.7a1 1 0 0 1 0-1.42l4.95-4.95a8 8 0 0 1 11.32 0l4.95 4.95a1 1 0 0 1 0 1.42l-4.95 4.95-.01.01zm-9.9-1.42a6 6 0 0 0 8.48 0L19.38 8l-4.24-4.24a6 6 0 0 0-8.48 0L2.4 8l4.25 4.24h.01zM10.9 12a4 4 0 1 1 0-8 4 4 0 0 1 0 8zm0-2a2 2 0 1 0 0-4 2 2 0 0 0 0 4z"></path>
```
emargareten, line 624: same as views icon

composer.json
```diff
@@ -16,6 +16,7 @@
     "require": {
         "php": "^7.3|^8.0",
         "ext-json": "*",
+        "guzzlehttp/guzzle": "^7.3",
```
barryvdh, line 19: I don't think Guzzle itself is required?

resources/views/layout.blade.php
```diff
@@ -197,6 +197,14 @@
                             <span>Views</span>
                         </router-link>
                     </li>
+                    <li class="nav-item">
+                        <router-link active-class="active" to="/client-requests" class="nav-link d-flex align-items-center">
+                            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20">
+                                <path d="M7 3H2v14h5V3zm2 0v14h9V3H9zM0 3c0-1.1.9-2 2-2h16a2 2 0 0 1 2 2v14a2 2 0 0 1-2 2H2a2 2 0 0 1-2-2V3zm3 1h3v2H3V4zm0 3h3v2H3V7zm0 3h3v2H3v-2z"/>
```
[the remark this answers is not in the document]
gdebrauwer: I replaced it with a globe icon from [zondicons](https://www.zondicons.com/icons.html). (that is also the source of the other icons according to https://github.com/laravel/telescope/pull/679#issuecomment-522022489)

composer.json
gdebrauwer: Locally I needed to install guzzle to run the tests, but you are right that this should not be a requirement to install Telescope. I removed the dependency and I updated the "tests" github workflow to install the "guzzlehttp/guzzle" composer package before running the tests.

.github/workflows/tests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-           composer require "illuminate/contracts=${{ matrix.laravel }}" --no-update
+           composer require "illuminate/contracts=${{ matrix.laravel }}" "guzzlehttp/guzzle" --no-update
```
driesvints, line 42: Add this to the dev requirements instead please.

driesvints:
Hey @gdebrauwer, can you also send in a PR to document this? https://laravel.com/docs/8.x/telescope

gdebrauwer:
Yes of course! 👍

PR #1089, opened 2021-07-08
Added a fix for JobWatchers when the job payload is encrypted

morloderex:
Hey guys. 

I have opened this pr because i encountered a problem when using the JobEncryption feature for a given job.

Telescope assumes the Job is already ready to be unserialized and does not handle encryption

The logic have i stolen from here:
https://github.com/laravel/framework/blob/108385b4f98cacfc1ef1d6e323f57b1c2df3180f/src/Illuminate/Queue/CallQueuedHandler.php#L86-L105

Therefore i have not provided any tests sense the logic is exactly the same apart from the container not being present on the instance itself.

### DIFF
--- a/src/Watchers/JobWatcher.php
+++ b/src/Watchers/JobWatcher.php
@@ -3,9 +3,11 @@
 namespace Laravel\Telescope\Watchers;
 
 use Illuminate\Bus\BatchRepository;
+use Illuminate\Contracts\Encryption\Encrypter;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Were you able to confirm this fixes the issue on your Telescope installation?

morloderex:
Yes.

After adding this code to the watcher it was working as expected.

I can add test if it’s needed but as the code is the same as in the framework I feel like it’s testing the same thing twice.

PR #1093, opened 2021-07-12
[4.x] Allow select entry types to be pruned

emargareten:
While pruning prevents accumulating records, there are cases when we do want to keep some of the data (e.g. exceptions that still needs to be resolved).

This PR allows for pruning selected entry types by adding `--type` option to the `prune` command.

### DIFF
--- a/src/Contracts/PrunableRepository.php
+++ b/src/Contracts/PrunableRepository.php
@@ -10,7 +10,8 @@ interface PrunableRepository
      * Prune all of the entries older than the given date.
      *
      * @param  \DateTimeInterface  $before
+     * @param  array  $entryTypes
      * @return int
      */
-    public function prune(DateTimeInterface $before);
+    public function prune(DateTimeInterface $before, array $entryTypes = []);
 }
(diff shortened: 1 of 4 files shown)


src/Contracts/PrunableRepository.php
```diff
@@ -10,7 +10,8 @@ interface PrunableRepository
      * Prune all of the entries older than the given date.
      *
      * @param  \DateTimeInterface  $before
+     * @param  array  $entryTypes
      * @return int
      */
-    public function prune(DateTimeInterface $before);
+    public function prune(DateTimeInterface $before, array $entryTypes = []);
```
driesvints, line 16: This is a breaking change.

src/Storage/DatabaseEntriesRepository.php
```diff
@@ -333,11 +333,15 @@ public function stopMonitoring(array $tags)
      * Prune all of the entries older than the given date.
      *
      * @param  \DateTimeInterface  $before
+     * @param  array  $entryTypes
      * @return int
      */
-    public function prune(DateTimeInterface $before)
+    public function prune(DateTimeInterface $before, array $entryTypes = [])
```
driesvints, line 339: This is a breaking change.

emargareten: @driesvints Why is it a breaking change?

driesvints: If anyone is implementing the contract they'll see their apps break.

emargareten:
Ha, I see master already has something similar https://github.com/laravel/telescope/commit/7bfe65fe17905eb2fb4836b0346e09eb04374739

PR #1097, opened 2021-07-21
[4.x] Record 'connection failed' http client event

gdebrauwer:
This makes it possible to record a Http Client request that did not receive a response.

I did not add any tests for it because I don't think it is possible to fake a connection exception using Http Client Fake 🤔


resources/js/mixins/entriesStyles.js
```diff
@@ -45,7 +45,7 @@ export default {
             if (status < 300) return 'secondary';
             if (status < 400) return 'info';
             if (status < 500) return 'warning';
-            if (status >= 500) return 'danger';
+            if (status >= 500 || !$status) return 'danger';
```
taylorotwell, line 48: `$status`? With a dollar sign? Is that correct? Did you test these changes?

gdebrauwer: Oh 🙈 I did test it and it worked just fine 😅 I will fix it

PR #1106, opened 2021-07-30
Update index.vue to show Host of request

alitokmakci:
For who using subdomains show the host of request path.

When using a subdomain, more than one request sent to the same path appears on the requests page. To clear up the confusion, a host column can be added to the table.

### DIFF
--- a/resources/js/screens/requests/index.vue
+++ b/resources/js/screens/requests/index.vue
@@ -12,6 +12,7 @@
     <index-screen title="Requests" resource="requests">
         <tr slot="table-header">
             <th scope="col">Verb</th>
+            <th scope="col">Host</th>
             <th scope="col">Path</th>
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to add this atm.

NickSdot:
@taylorotwell I am feeling the same pain in multiple spots.

Would you be open to a PR that adds **no new table columns**, but a config option `domain_aware` with `true` / `false`, with false as default? Instead of adding new table columns, we simply record the `fullUrl()` if the user wishes so.

Something like this:

```php
Telescope::recordRequest(IncomingEntry::make([

    // ...

    'uri' => config('telescope.domain_aware')
        ? $event->request->fullUrl()
        : (str_replace($event->request->root(), '', $event->request->fullUrl()) ?: '/'),

    // ...

]));
```

Issue #1023, opened 2021-03-04
Connecting to Redis during package discovery


rihardssceredins:
- Telescope Version: v4.4.4
- Laravel Version: v8.29.0
- PHP Version: 8.0.2
- Database Driver & Version:

### Description:
Hey,

Is it by design that Telescope is trying to connect to Redis during package discovery when running `composer install`? 🤔
I'm using docker-compose for my local environment and it's really driving me nuts that I've to start all services before I can install composer packages or switch a branch (I've a post checkout & post merge git hook to run `composer install` automatically). 😬

<img width="1107" alt="Screenshot 2021-03-04 at 09 46 05" src="https://user-images.githubusercontent.com/23099574/109929536-3d182880-7ccf-11eb-869a-e2f5c56e086e.png">

driesvints:
Can you post a full stacktrace so we can pinpoint the line where Telescope is trying to connect to Redis?

rihardssceredins:
@driesvints sorry, didn't think of it before. Here you go:

```
[2021-03-04 11:16:57] local.ERROR: Connection refused {"exception":"[object] (RedisException(code: 0): Connection refused at /Users/rihardssceredins/Code/api-diet/vendor/laravel/framework/src/Illuminate/Redis/Connectors/PhpRedisConnector.php:142)
[stacktrace]
#0 /Users/rihardssceredins/Code/api-diet/vendor/laravel/framework/src/Illuminate/Redis/Connectors/PhpRedisConnector.php(142): Redis->connect('127.0.0.1', 6379, 0, NULL, 0, 0)
#1 /Users/rihardssceredins/Code/api-diet/vendor/laravel/framework/src/Illuminate/Redis/Connectors/PhpRedisConnector.php(84): Illuminate\\Redis\\Connectors\\PhpRedisConnector->establishConnection(Object(Redis), Array)
#2 /Users/rihardssceredins/Code/api-diet/vendor/laravel/framework/src/Illuminate/Support/helpers.php(263): Illuminate\\Redis\\Connectors\\PhpRedisConnector->Illuminate\\Redis\\Connectors\\{closure}(Object(Redis))
#3 /Users/rihardssceredins/Code/api-diet/vendor/laravel/framework/src/Illuminate/Redis/Connectors/PhpRedisConnector.php(109): tap(Object(Redis), Object(Closure))
[...]

PR #1025, opened 2021-03-08
[4.x] Move cache read inside handler to avoid early cache call

rihardssceredins:
Move cache read inside VarDumper handler to avoid cache call during package discovery/composer install calls. Fixes #1023.

PR #1116, opened 2021-08-19
[4.x] Add cache driver option

wsamoht:
# Summary
This PR adds a `cache_driver` option to the `telescope.php` config in order to override the cache store/driver that Telescope uses throughout.

# Motivation
The motivation of this is due to the error in #1023. When you are developing locally using a redis cache which is not local due to using Homestead, Sail, or some other virtual environment, `composer` and `artisan` commands ran locally fail due to the redis cache not being accessible. The redis config in the `.env` cannot be changed as it is needed for the virtual environment - meaning we cannot use the forwarded port as the virtual environment needs the actual port (e.g. `6379` and not `63790`).

I personally ran into the issue trying to run `composer update`, `php artisan make:test` and even just `php artisan`.

# Not a breaking change
This PR is backward compatible. If the `cache_driver` option doesn't exist, it defaults to the default cache driver.

# Remaining tasks
- [ ] Improve the description for the option in the `telescope.php` config
- [ ] Write some tests

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -53,6 +53,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Telescope Cache Driver
+    |--------------------------------------------------------------------------
+    |
+    | This configuration options determines the cache driver that will be
+    | used throughout Telescope.
+    | Hmmmm...
+    |
+    */
+
+    'cache_driver' => env('TELESCOPE_CACHE_DRIVER', 'default'),
+
+    /*
+    |--------------------------------------------------------------------------
     | Telescope Master Switch
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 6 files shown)


taylorotwell:

wsamoht:
@taylorotwell thank you for reviewing. Is there a different way of handling this issue that would be more acceptable? I'd be glad to see if I can figure out a different solution. Per @driesvints in [this](https://github.com/laravel/telescope/pull/1025#issuecomment-810065398) comment, a PR would be appreciated.

I do still believe it is a bug as it makes using Telescope unusable or local commands when using a non-local redis instance. Memcached and the database drivers also have this same issue.

driesvints:
@wsamoht I guess Taylor doesn't wants to pursue a fix at this time, sorry.

PR #1118, opened 2021-08-25
Add auto night theme mode

joowdx:
Using media(prefers-color-scheme: dark) in link stylesheet so theme can automatically switch between light and dark mode on the fly since current telescope's night theme setting doesn't follow the operating system's or the browser's settings.


driesvints:
I think it might be better if we adopted the same JS approach that was shipped to the docs: https://github.com/laravel/laravel.com-next/pull/180

We could also then see if we could use that for Horizon.

taylorotwell:
See above.

PR #1119, opened 2021-08-26
Always register migrations

MasterRO94:
I think we need to register migrations even if telescope is disabled so database would be ready when we switch state enable/disable. So there is no need to check and run migrations every time we enable telescope. It can be useful for CI/CD.


lk77:
@MasterRO94 was this reverted ? 

it does not seems to be the case anymore

MasterRO94:
Yes, we use my PR patch via composer patches plugin https://github.com/cweagans/composer-patches
Don't know why master branch is not maintained

PR #1123, opened 2021-09-09
[3.x] Fix `telescope.domain` usage.

finagin:
```php
// config/telescope.php
return [
    'domain' => 'telescope.example.com',
    'path' => null,
];
```
If you use `domain` without `path` `handlingApprovedRequest` always return false because `config('telescope.path').'*'` is `'*'` 
I propose check domain before check path.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -186,7 +186,36 @@ class Telescope
      */
     protected static function handlingApprovedRequest($app)
     {
-        return ! $app->runningInConsole() && ! $app['request']->is(
+        if ($app->runningInConsole()) {
+            return false;
+        }
+
+        return static::requestIsToApprovedDomain($app['request'])
+            || static::requestIsToApprovedUri($app['request']);
+    }
+
+    /**
+     * Determine if the request is to an approved domain.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return bool
+     */
+    protected static function requestIsToApprovedDomain($request): bool
+    {
+        $currentHost = $request->getHost();
+
+        return config('telescope.domain', $currentHost) !== $currentHost;
+    }
+
+    /**
+     * Determine if the request is to an approved URI.
+     *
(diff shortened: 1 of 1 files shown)


src/Telescope.php
```diff
@@ -186,7 +186,36 @@ protected static function runningApprovedArtisanCommand($app)
      */
     protected static function handlingApprovedRequest($app)
     {
-        return ! $app->runningInConsole() && ! $app['request']->is(
+        if ($app->runningInConsole()) {
+            return false;
+        }
+
+        return static::requestIsToApprovedDomain($app['request'])
+            || static::requestIsToApprovedUri($app['request']);
```
taylorotwell, line 194: Why is this an "or" condition?

taylorotwell:
Only needs to be sent to current branch.

finagin:
> Only needs to be sent to current branch.

Hey @taylorotwell!
I ran into this problem on a project with laravel 6 and I need these changes there because we cannot update laravel yet.
These changes don't break anything

src/Telescope.php
finagin: Because if this is a separate domain for telescope, we should not keep a record, if the domain is not separate, we check the paths

driesvints:
@finagin we only maintain the latest versions of our packages. Please upgrade to Laravel 8

finagin:
> @finagin we only maintain the latest versions of our packages. Please upgrade to Laravel 8

Laravel 6 is LTS right?
Branch telescope 3.x exists?
I think there is no problem to merge these changes

driesvints:
Laravel 6 is LTS but the first party packages aren't. You can find all that info in our support policy https://laravel.com/docs/8.x/releases#support-policy

finagin:
Sorry, but then I don't understand why the 3.x branch is needed instead of 3.0? Here and at Horizon. 
Tell me if it's not difficult.

PR #1125, opened 2021-09-14
Http Client Watcher Does not Log XML Request and Responses

4nik3t:
Here is the before request and response display of XML Requests in Telescope HTTP Client Watcher

**Before**
image

**After**
image


taylorotwell:
No plans to change this right now. It isn't formatted very nicely.

Issue #1126, opened 2021-09-14
telescope_entries table queries are shown in Query Watcher


ismaail:
- Telescope Version: v4.6.3
- Laravel Version:  v8.61.0
- PHP Version: 8.0.10
- Database Driver & Version: MySQL 8

### Description:
`telescope_entries` table queries are shown in Query Watcher.

https://user-images.githubusercontent.com/4411670/133323960-38047e8f-2556-42af-9e59-f4a8fa4df2ad.mp4

### Steps To Reproduce:
1. Upgrade Laravel Telescope to version 4.6.3
2. Have `TELESCOPE_DOMAIN` in the `.env` file with empty value or remove it.

RTippin:
I can confirm having this behavior as well after updating to `v4.6.3`. Even my request watcher was showing all telescope `POST` to the internal `telescope-api`.

metalogico:
Yes, same problem for me. I just left a comment on the commit which is likely to cause the issue.

SaaberDev:
facing same issue .... 😫😫

ismaail:
@SaaberDev workaround: add `TELESCOPE_DOMAIN` to `.env` with the app domain as its value.

SaaberDev:
@ismaail 

![image](https://user-images.githubusercontent.com/60372951/133658127-29c0a4d2-2edc-48ea-86b7-0d826db72cd0.png)


Not working ....

ismaail:
@SaaberDev try without `https://`

SaaberDev:
@taylorotwell .... Thanks boss .. Got the new update just now and fixed ❤❤

PR #1127, opened 2021-09-15
[4.x] Don't log HTTP requests to Telescope endpoints

derekmd:
Fixes https://github.com/laravel/telescope/issues/1126 introduced by https://github.com/laravel/telescope/pull/1124

Taylor rightly questioned [why the domain check was an OR conditional](https://github.com/laravel/telescope/pull/1123#discussion_r705621841) as that ended up breaking Telescope's behavior using the out-of-the-box config:

```php
return [
    'domain' => env('TELESCOPE_DOMAIN', null),
    'path' => env('TELESCOPE_PATH', 'telescope'), // e.g., https://example.com/telescope
    // ...
];
```

With Telescope's default `config('telescope.domain') === null`, the new method `requestIsToApprovedDomain()` was always returning true. This skips `requestIsToApprovedUri()` and permits every path to be logged. Telescope database entries are being inserted for /telescope/requests, /telescope/batches, etc. while also ignoring `config('telescope.only_paths')`.

Instead make changes that allow the below config to be used, which the pull request originally intended to fix:

```php
return [
    'domain' => 'https://telescope.example.com',
    'path' => null,
    // ...
];
```

1. `requestIsToApprovedDomain()` must check `config('telescope.domain')` contains a value for Telescope route registration. A request is approved for logging when Telescope's domain configuration is empty or its filled value is different from the current request meaning it's an app request (and not a Telescope admin request.)

   > The issue with https://github.com/laravel/telescope/pull/1124 was its fallback default value of `config('telescope.domain', $request->getHost())` only gets used when 'domain' array key _doesn't exist_. config/telescope.php assigns that key the value `null` and fallback `$request->getHost()` was never returned.
2.  `requestIsToApprovedUri()` will ignore `config('telescope.path') === null` for custom domains since conditional `! $request->is('*')` would ignore _all_ paths and nothing is ever logged.
3. When `requestIsToApprovedDomain()` returns `true`, `config('telescope.only_paths')` must also be checked. So that pull request should have introduced an AND conditional. The fixes in bullet points 1 & 2 allow  `requestIsToApprovedDomain() && requestIsToApprovedUri()` to be used.

I wouldn't mind adding some integration and/or unit tests for this but any Laravel app code skipped by `! $app->runningInConsole()` (like protected method `handlingApprovedRequest()`) is untestable with PHPUnit unless the code is restructured to make that mockable.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -200,8 +200,8 @@ class Telescope
             return false;
         }
 
-        return static::requestIsToApprovedDomain($app['request'])
-            || static::requestIsToApprovedUri($app['request']);
+        return static::requestIsToApprovedDomain($app['request']) &&
+            static::requestIsToApprovedUri($app['request']);
     }
 
     /**
@@ -212,9 +212,8 @@ class Telescope
      */
     protected static function requestIsToApprovedDomain($request): bool
     {
-        $currentHost = $request->getHost();
-
-        return config('telescope.domain', $currentHost) !== $currentHost;
+        return is_null(config('telescope.domain')) ||
+            config('telescope.domain') !== $request->getHost();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


frostfire64:
I wasted half a day working on this exact issue in a project I'm working on. I think this merge request should me merged ASAP because it can break the behavior of Telescope config for any user working with it that updates the package.

I can confirm that it broke the default *and* documented behavior of telescope. Which generated major confusion for me since it was a clean install of telescope done yesterday.

frostfire64:
I tested this on the newest release, the issue is solved. Thanks for swift action 💪

EldonYoder:
Working great for me as well... thanks guys!

Issue #851, opened 2020-03-25
All Mail fails to display if any entry has a null to address


hackel:
- Telescope Version: 3.2.0
- Laravel Version: 6.18.2
- PHP Version: 7.4.3
- Database Driver & Version: MySQL 8

### Description:
I accidentally sent an email without a to address.  This was recorded correctly in Telescope, however it prevents the entire list of emails from displaying.  When loading the Mail index, it just shows "Scanning..." indefinitely.  No JS console or other errors are shown.

If I check the HTTP response, the call to `/telescope/telescope-api/mail?tag=&before=&take=50&family_hash=` returns a valid response with all of my mail entries.  I started going through and removing the old mail telescope_entries until it worked again, and the discovered that the offending record contained `content->to = null`.

### Steps To Reproduce:
Create a new Notification, and return a Mailable from the `toMail` method, but fail to call `to()` on it to specify a recipient address.

driesvints:
Telescope didn't record the mail for me. If you can add very specific sets to reproduce this we'll have another look. Thanks.

PR #1133, opened 2021-10-01
Fix mail to check to allow for null `to` fields

swarley:
We have notifications that are sent using solely `bcc` recipients in our app and this causes the `Mail` resource to load infinitely.

This change accounts for the fact that `entry.content.to` can be null for a valid mailable.

Related #851.

### DIFF
--- a/resources/js/screens/mail/index.vue
+++ b/resources/js/screens/mail/index.vue
@@ -2,7 +2,7 @@
     export default {
         methods: {
             recipientsCount(entry){
-                return _.union(Object.keys(entry.content.to),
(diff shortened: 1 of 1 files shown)


resources/js/screens/mail/index.vue
```diff
@@ -2,7 +2,7 @@
     export default {
         methods: {
             recipientsCount(entry){
-                return _.union(Object.keys(entry.content.to),
+                return _.union((entry.content.to ? Object.keys(entry.content.to) : []),
```
taylorotwell, line 5: Is there an extra `(` here?

PR #1155, opened 2021-12-01
Display `Run migrations first` when entries table is missing

faissaloux:
Hello :wave:

The issue we have here is that `Scanning...` message and the icon keeps spinning without any indication even if we have `entries table missing` as you can see in pictures below.

Screenshot from 2021-12-01 21-21-08
Screenshot from 2021-12-01 21-24-40

So in this PR I'm showing a message that reminds to run migrations first  `Run migrations first`.

Screenshot from 2021-12-01 21-36-03


taylorotwell:

faissaloux:
@taylorotwell thank you but I didn't add so much code that could break anything.

I'm just showing a reminder message that could save time waiting for the non-stop "scanning".

PR #1160, opened 2021-12-07
Allow filtering many tags

alancolant:
Adds the ability to filter entries with multiple tags separated by comma.
Example: status:200,graphql will only display entries with these 2 tags (status:200 and graphql)


driesvints:
Please undo all styling changes and resubmit. Thanks

PR #1166, opened 2022-01-06
Speed up "Clear Entries" action

nazmulpcc:
This PR disables foreign key checks while clearing telescope entries. Also "truncate" is used instead of "delete" which is significantly faster. This will improve large data deletion.


taylorotwell:
No plans to change this behavior atm.

nazmulpcc:
May I ask the reason behind a slower approach for learning purpose?

Issue #626, opened 2019-05-08
Exception entries do not get stored when a job fails


benignant:
- Telescope Version: 2.0.4
- Laravel Version: 5.8.16
- PHP Version: 7.3.3
- Database Driver & Version: MySQL 5.7.19

### Description:
Exception entries do not get stored when a job fails. The exception does however get stored as part of the job entry. Is this the expected behavior? It seems inconsistent since when a job triggers an exception but does not fail, the exception entry does get logged. I would propose that exception entries should get stored regardless of whether or not the job fails. The exception would then be linked to the job via the `batch_id`, so you wouldn't even need to store the exception in the job entry. Linking the exception to the job would also have the added benefit of indicating that a pending or processed job has already been attempted and resulted in an exception. Currently there is no way to tell that a job had to be attempted multiple times.

### Steps To Reproduce:

1. Process a job that triggers an exception but does not fail (because it will be retried)
2. Now process the job again, but this time ensure that it fails
3. Note that the first time the job was processed an exception entry was stored but not the second time

themsaid:
Can't replicate, works for me. Also doesn't seem like other people are reporting the same issue so I believe you'll need to look into your project code.

taylorotwell:
Can't recreate this either. Closing for now.

antoinelame:
I'm posting this message a few years later, for those like me who face the same issue.

The behavior described by @benignant is still actual and I see it on all the Laravel projects I work on.

Exceptions triggered by jobs **do not appear** in the Exceptions page of Telescope (`/telescope/exceptions`). However, they are showed on the detail page of a job that failed (`/telescope/jobs/{uuid}`). I don't know if this is the expected behavior or not.

This is because in `Laravel\Telescope\Watchers\ExceptionWatcher@recordException`, the `Telescope::isRecording()` value is equals to `false`.

```php
/**
 * Record an exception was logged.
 *
 * @param  \Illuminate\Log\Events\MessageLogged  $event
 * @return void
 */
public function recordException(MessageLogged $event)
{
    if (! Telescope::isRecording() || $this->shouldIgnore($event)) {
        return;
    }

    // ...
}
```

That said, this could be the expected behavior.

mshamaseen:
Returning to this after 2 years, isn't this a weird behavior? Shouldn't Telescope log job exceptions by default?

PR #1173, opened 2022-01-27
[4.x] Add exception context on job details

antoinelame:
When an exception is thrown in a job, it is not shown on the "Exceptions" page but only on the "Job Details" page ([see this discussion](https://github.com/laravel/telescope/issues/626#issuecomment-1023507962)). The context of an exception coming from a job is therefore never shown on Telescope.

This PR remedies this by adding a "Context" tab on failed job pages.

Screenshot

### DIFF
--- a/src/Watchers/JobWatcher.php
+++ b/src/Watchers/JobWatcher.php
@@ -115,7 +115,7 @@ class JobWatcher extends Watcher
                     'line_preview' => ExceptionContext::get($event->exception),
                     'context' => method_exists($event->exception, 'context')
                         ? $event->exception->context()
-                        : [],
+                        : null,
                 ],
             ]
         )->addTags(['failed']));
--- a/src/Watchers/JobWatcher.php
+++ b/src/Watchers/JobWatcher.php
@@ -113,6 +113,9 @@ class JobWatcher extends Watcher
                     'trace' => $event->exception->getTrace(),
                     'line' => $event->exception->getLine(),
                     'line_preview' => ExceptionContext::get($event->exception),
+                    'context' => method_exists($event->exception, 'context')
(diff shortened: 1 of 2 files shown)


src/Watchers/JobWatcher.php
```diff
@@ -113,6 +113,9 @@ public function recordFailedJob(JobFailed $event)
                     'trace' => $event->exception->getTrace(),
                     'line' => $event->exception->getLine(),
                     'line_preview' => ExceptionContext::get($event->exception),
+                    'context' => method_exists($event->exception, 'context')
```
taylorotwell, line 116: Is this some new convention? Exceptions don't usually have a `context` method I don't think?

antoinelame: Isn't it a Laravel convention?

The [documentations says](https://laravel.com/docs/8.x/errors#exception-log-context) the data passed in the `context()` method is automatically included in logs. And [Telescope already shows](https://github.com/laravel/telescope/blob/4.x/resources/js/screens/exceptions/preview.vue#L90) the `context()` data in the Exception page (this PR is for adding the same to the Failed Job page).

taylorotwell:
But that isn't how the context is retrieved on the backend by the exception watcher. Exceptions do not have a "context" method. Did you test this new screen in a full application using Telescope?

antoinelame:
@taylorotwell I replicated [this method](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Exceptions/Handler.php#L291) that is found in the exception Handler. How should I do otherwise to display the context on this screen? Having the context not only in the logs but also on Telescope is really handy. 🤙

Yes I tested in full applications. I actually use this fork on several applications since yesterday.

Issue #1081, opened 2021-06-23
Telescope on vapor issue


denjaland:
- Telescope Version: 4.5.1
- Laravel Version: 8.48.0
- PHP Version: vapor php-8.0:al2
- Database Driver & Version:

### Description:
To support multi tenancy, whenever we dispatch a queued job, we are adding some payload information on it.
One of these details is a reference to the tenant that the user was working on when the jobs gets dispatched:

```php
$this->app['queue']->createPayloadUsing(function(string $connection, string $queue = null, array $payload = []) {
  return array_merge($payload, ['tenant' => optional(Tenancy::tenant())->id()]);
});
```

Whenever a job is being picked up from the queue, we have the following running:
```php
$this->app['events']->listen(JobProcessing::class, function($event) {
  $tenantId = Arr::get($event->job->payload(), 'tenant');
  if($tenantId) {
    Tenancy::setTenant(Tenant::find($tenantId));
  }
});
```

This basically allows us - when a user does a password reset for instance on any of the tenants - we know what website he was visiting, and can nicely theme our outgoing emails to match the website he was visiting.

This works great, and we have already debugged that Tenant::find($tenantId) does in fact give me the correct tenant.
The functioanlity also works perfectly as long as we don't have the Telescope ModelWatcher activated.

As soon as we do, however, I'm getting an error returned by Telescope:
`Cannot access offset of type string on string`

It is thrown by `/var/task/vendor/laravel/telescope/src/Watchers/ModelWatcher.php` on line 87:

```php
public function recordHydrations($data)
    {
        if (! ($this->options['hydrations'] ?? false)
            || ! $this->shouldRecordHydration($modelClass = get_class($data[0]))) {
            return;
        }

        if (! isset($this->hydrationEntries[$modelClass])) {
            $this->hydrationEntries[$modelClass] = IncomingEntry::make([
                'action' => 'retrieved',
                'model' => $modelClass,
                'count' => 1,
            ])->tags([$modelClass]);

            Telescope::recordModelEvent($this->hydrationEntries[$modelClass]);
        } else {
            $this->hydrationEntries[$modelClass]->content['count']++;  // THIS LINE THROWS THE EXCEPTION
        }
    }
```

At this time I only seem to be able to reproduce this when deployed on vapor, but I'm suspecting it might be a more generic issue with php 8+, but not really sure.

Looking at the code, I don't really know why this error can occur either, but maybe someone here has an idea what might be causing this odd behaviour?

[...]

Issue #1068, opened 2021-06-01
Illegal string offset 'count' under ModelWatcher


daison12006013:
- Telescope Version: v4.4.3
- Laravel Version: 8.14.0
- PHP Version: v7.4.16
- Database Driver & Version: MySQL Ver 14.14 Distrib 5.7.33-36, for debian-linux-gnu (x86_64) using  7.0

### Description:
I was trying to simulate the telescope thru test cases (ton of test cases functional/unit) with telescope enabled, moreover I am getting this error 

```
[ErrorException] Illegal string offset 'count'

#1  /var/www/xxxxx/app/Services/Telescope/Watchers/ModelWatcher.php:42
#2  /var/www/xxxxx/vendor/laravel/telescope/src/Watchers/ModelWatcher.php:49
#3  /var/www/xxxxx/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:388
#4  /var/www/xxxxx/vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:236
#5  /var/www/xxxxx/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php:189
#6  /var/www/xxxxx/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Model.php:434
#7  /var/www/xxxxx/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:324
#8  Illuminate\Database\Eloquent\Builder->Illuminate\Database\Eloquent\{closure}
#9  /var/www/xxxxx/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:323
#10 /var/www/xxxxx/vendor/laravel/framework/src/Illuminate/Support/Traits/ForwardsCalls.php:23
```

I came across debugging the `IncomingEntry.php` and for some reason the property `$content` turns out to be a json!? weird!

![image](https://user-images.githubusercontent.com/4581415/120289223-a53cd180-c2f3-11eb-835f-263474643783.png)

Here's how I dumped above
![image](https://user-images.githubusercontent.com/4581415/120289357-cdc4cb80-c2f3-11eb-8406-0f90238fdc60.png)

### Steps To Reproduce:
it's not a roadblock when using telescope thru manual testing, but it's really noisy in our Sentry / Kibana, reproducing it seems really hard, because the initial state its an array then some of the hydrated entries became json string.

### Work Around:
[...]

PR #1000, opened 2020-12-12
[4.x] Add model hydrations to model watcher

paras-malhotra:
### Motivation
Re-attempt of https://github.com/laravel/telescope/pull/999 without a separate tab or watcher.

This PR adds the ability to record model hydrations within the Model Watcher (without a separate tab or watcher). This would help to know how many models are being hydrated for a request - something that can have a significant on performance / memory usage. Inspired from a similar popular feature of Debugbar: barryvdh/laravel-debugbar#947.

There are no breaking changes with this PR. To enable recording of model hydrations, we would need to enable the `hydrations` flag in the `config/telescope.php` file like so:

```php
Watchers\ModelWatcher::class => [
    'enabled' => env('TELESCOPE_MODEL_WATCHER', true),
    'events' => ['eloquent.*'],
    'hydrations' => true,
],
```

### Index Screen (The retrieved is the one for model hydrations)
hydrations

### Preview Screen
hydrations-preview

### Related Entries
hydrations-related-entries

### Note
This PR would need a re-build and I haven't committed the compiled files (just the source) as per the contribution guidelines.

PR #1174, opened 2022-01-31
[4.x] Attempt to fix model hydration

driesvints:
This is an attempt to fix https://github.com/laravel/telescope/issues/1081 & https://github.com/laravel/telescope/issues/1068.

https://github.com/laravel/telescope/pull/1000 introduced the ability to record how many hydrations of a specific model occurred. This however, does not seem to function properly on Vapor specifically. [It was originally also post-reported on the related PR](https://github.com/laravel/telescope/pull/1000#discussion_r559066549) but given the nature of it being Vapor specific it was very hard to reproduce for the OP.

My analysis leads me to believe that the saving of the incoming entry occurs and then afterwards, for some reason unknown, Telescope still attempts to hydrate a model when being run on Vapor. Because [the incoming entry's `content` has already been json encoded here](https://github.com/laravel/telescope/blob/4.x/src/Storage/DatabaseEntriesRepository.php#L145), it's now a string and the following exception occurs for people:

> Cannot access offset of type string on string

I think this is because of some processing that happens after recording the entry which hydrates the same model again. It seems Vapor specific but most likely it's specifically due to any way an app handles processing after Telescope recording. However I could not in any way reproduce this locally.

Therefor I'm not 100% sure this fix will provide a solution to the aforementioned issues. Someone managed to solve it [by modifying the `DatabaseEntriesRepository`](https://github.com/laravel/telescope/issues/1068#issuecomment-852019407) but I don't think that's the correct place since this is specific to the `ModelWatcher`. I do not entirely know if Telescope will attempt to save the new entries again or if it'll ignore it.

I'd appreciate it if anyone from the related issues could attempt to deploy a fork of Telescope with this to see if that solves the issue for them on Vapor.

### DIFF
--- a/src/Watchers/ModelWatcher.php
+++ b/src/Watchers/ModelWatcher.php
@@ -51,15 +51,15 @@ class ModelWatcher extends Watcher
             return;
         }
 
-        $model = FormatModel::given($data[0]);
+        $modelClass = FormatModel::given($data[0]);
 
         $changes = $data[0]->getChanges();
 
         Telescope::recordModelEvent(IncomingEntry::make(array_filter([
             'action' => $this->action($event),
-            'model' => $model,
+            'model' => $modelClass,
             'changes' => empty($changes) ? null : $changes,
-        ]))->tags([$model]));
+        ]))->tags([$modelClass]));
     }
 
     /**
@@ -84,7 +84,13 @@ class ModelWatcher extends Watcher
 
             Telescope::recordModelEvent($this->hydrationEntries[$modelClass]);
         } else {
-            $this->hydrationEntries[$modelClass]->content['count']++;
+            $entry = $this->hydrationEntries[$modelClass];
+
+            if (is_string($this->hydrationEntries[$modelClass]->content)) {
+                $entry->content = json_decode($entry->content, true);
+            }
+
+            $entry->content['count']++;
         }
     }
 


daison12006013:
Hi from #1068,

Based on your changes, it is similar to what I did.

> My work around was to extend the class and modify the config to use this class and I am checking first if the content was string then just json_decode it to be an array.

Perhaps that should work, like I said before this is somewhat a peculiar case that doesn't occur all the time.

taylorotwell:
The only way I see that this could happen is if:

A) `startRecording` is somehow called during the `withoutRecording` callback in `Telescope@store`. 

B) The after storage hooks are not being invoked because of some exception being thrown within `Telescope@store` before the after hooks can be invoked at the end of the callback.

The `ModelWatcher` class registers an "after storage" hook that flushes all of its `hydrationEntries`. The only place those are serialized to JSON strings is during the database storage process. So, somehow a model hydration event is being recorded within the `withoutRecording` callback, after storage has happened, but before the after storage callbacks have been invoked.

It's hard for me to imagine how that could be happening but I guess it is somehow? A and B noted above are the only situations I can think of that could cause this but we would need to probably place some logging statements on an application running on Vapor to try to diagnose which one is happening.

driesvints:
@taylorotwell right. In that case we'd need to ask one of the involved people to adjust Telescope to place the logging.

taylorotwell:
Will merge this for now but would still like to discover the root cause of why this is happening if someone is willing to put some debug statements in place.

Issue #1178, opened 2022-02-01
ERROR: ErrorException: Undefined array key 0


yasirhashmi88:
- Telescope Version: 4.7.1
- Laravel Version: 8
- PHP Version: 8
- Database Driver & Version: MySQL 8.0.23

### Description:

Getting `ERROR: ErrorException: Undefined array key 0 in /var/www/hr/vendor/laravel/telescope/src/Watchers/ModelWatcher.php:74`

- If I set `'hydrations' => false,` in `telescope.php`, it works.
- I noticed the key of an array comes as string like `model` and doesn't comes as `0`

### Steps To Reproduce:

1 - I have existing Laravel app which have different models but I believe this can be replicated on fresh Laravel installation.
2 - Run `composer require laravel/telescope:4.7.1 --dev`. I have tried multiple versions but still no luck.
3 - Try open your app in browser. You should see the error in the console or your logs.

driesvints:
I think this is related to my PR here: https://github.com/laravel/telescope/pull/1174. Seems that associative keys are kept because of the decoding. But it's beyond me why they're being recorded again.

driesvints:
@yasirhashmi88 can you try out https://github.com/laravel/telescope/pull/1180 ?

yasirhashmi88:
> @yasirhashmi88 can you try out #1180 ?

@driesvints Yes. That's fixes the issue.

PR #1180, opened 2022-02-03
[4.x] Fix key handling for ModelWatcher

driesvints:
It seems a breakage snuck in with my previous PR: https://github.com/laravel/telescope/pull/1174. Because the payload is decoded as an associative array (as should because we can't guarantee the data order in the payload), when new events happen with this payload we'll need to reference the named key instead. 

It's beyond me why new events are recorded but I guess this can happen since every eloquent model event is being watched. 

Fixes https://github.com/laravel/telescope/issues/1178

### DIFF
--- a/src/Watchers/ModelWatcher.php
+++ b/src/Watchers/ModelWatcher.php
@@ -51,9 +51,9 @@ class ModelWatcher extends Watcher
             return;
         }
 
-        $modelClass = FormatModel::given($data[0]);
+        $modelClass = FormatModel::given($data['model'] ?? $data[0]);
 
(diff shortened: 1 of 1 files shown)


yasirhashmi88:
@driesvints If you can, could you please create new release. Thanks

driesvints:
@yasirhashmi88 did this work for you?

driesvints:
Nvm. I'll tag a new release tomorrow.

yasirhashmi88:
@driesvints Yes it does. Thanks.

driesvints:
@yasirhashmi88 released.

PR #1199, opened 2022-03-08
Add option to record dumps even though the dump watcher tab is not open

superDuperCyberTechno:
This PR adds an environment variable (`TELESCOPE_DUMP_WATCHER_ALWAYS`) to let the user decide if Telescope should record dumps even if the dump watcher tab is closed.

I always expect my dumps to be recorded by Telescope and every time I forget to keep the dump watcher tab open, it throws me off. Here is a non-breaking solution to that.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -124,7 +124,10 @@ return [
             'ignore' => [],
         ],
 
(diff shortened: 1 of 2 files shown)


driesvints:
This seems to break the test suite. Please fix the tests and mark as ready when you're done.

superDuperCyberTechno:
My bad. On it.

Issue #655, opened 2019-06-24
[Feature request] Telescope pause/resume artisan command


rmundel:
Hi!

I was thinking that maybe some artisan commands to reflect the pause/resume feature from the Web UI could be useful.
I just encountered a situation where it would be nice to have the ability to pause telecope != business hours and resume otherwise.
Maybe:
php artisan telescope:pause or stop
php artisan telescope:resume or start

Thanks!

ManojKiranA:
looking pretty good to me

taylorotwell:
This is not a bug report so I am closing for now.

PR #1200, opened 2022-03-09
Add pause and resume commands

jamespavett:
Opening a pull request for the above based on closed issue #655, adding two new commands for pausing and resuming telescope.

Think it would be good functionality to have, and shouldn't break any existing tests as the commands shouldn't collide with anything. As the state of pause/resume is being stored in the cache, I have noticed times where post build/release of my own apps, we have cleared the cache and Telescope has started watching again, slowing down the production application, so this has become a manual activity to pause it again. Adding these commands would allow them to be triggered without the need for manual input.

Sorry if there is anything I'm missing, first time I've done this.

### DIFF
--- a/src/Console/PauseCommand.php
+++ b/src/Console/PauseCommand.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Cache\Repository as CacheRepository;
+
+class PauseCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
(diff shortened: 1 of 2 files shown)


jamespavett:
It's still showing as 1 StyleCI issue, but not sure why as I just pushed change to fix

taylorotwell:
Both of the commands were in the `App` namespace for some reason and neither one was actually registered in the service provider. I have corrected these things and made sure these actually work properly when installed in a Laravel application.

jamespavett:
Thank you! Yea those must have been oversight when migrating them over. I have been using them as my own custom commands for a while.

PR #1206, opened 2022-03-21
CacheWatcher hide values

dododedodonl:
When using telescope, I ran into a problem that cache values were too big to store (`SQLSTATE[08S01]: Communication link failure: 1153 Got a packet bigger than 'max_allowed_packet' bytes`). I do not need to know the value, but I do want to know the event happened. This feature allows to hide the value, based on the key and allowing wildcards.

### DIFF
--- a/src/Watchers/CacheWatcher.php
+++ b/src/Watchers/CacheWatcher.php
@@ -103,6 +103,33 @@ class CacheWatcher extends Watcher
     }
 
     /**
+     * Determine the value of an event.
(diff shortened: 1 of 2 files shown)


suyar:
It is expected that the changes to the configuration file will be noted at the same time as the release.  Just like this [commit](https://github.com/laravel/telescope/commit/9b1f19c2ef5bb396c25d4a204abcbfe29455d5b8).

PR #1207, opened 2022-03-22
Added option to skip exception and relative occurrences when prune

gennarobotta:
Hi, I have made a PR to give to possibility to keep exception and occurrences related to them, the reason is because in my case i have to make a prune every often but i would like to keep the exception and check them when i have time. This PR shouldn't break any existing features

### DIFF
--- a/src/Contracts/PrunableRepository.php
+++ b/src/Contracts/PrunableRepository.php
@@ -10,7 +10,7 @@ interface PrunableRepository
      * Prune all of the entries older than the given date.
      *
      * @param  \DateTimeInterface  $before
-     * @param  boolean  $skipException
+     * @param  bool  $skipException
      * @return int
      */
     public function prune(DateTimeInterface $before, $skipException);
--- a/src/Contracts/PrunableRepository.php
+++ b/src/Contracts/PrunableRepository.php
@@ -10,7 +10,8 @@ interface PrunableRepository
      * Prune all of the entries older than the given date.
      *
      * @param  \DateTimeInterface  $before
+     * @param  boolean  $skipException
      * @return int
      */
-    public function prune(DateTimeInterface $before);
+    public function prune(DateTimeInterface $before, $skipException);
 }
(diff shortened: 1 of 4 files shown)


src/Contracts/PrunableRepository.php
```diff
@@ -10,7 +10,8 @@ interface PrunableRepository
      * Prune all of the entries older than the given date.
      *
      * @param  \DateTimeInterface  $before
+     * @param  boolean  $skipException
      * @return int
      */
-    public function prune(DateTimeInterface $before);
+    public function prune(DateTimeInterface $before, $skipException);
```
driesvints, line 16: We can't add new arguments to methods or contracts on minor/patch releases.

src/Storage/DatabaseEntriesRepository.php
```diff
@@ -333,13 +333,20 @@ public function stopMonitoring(array $tags)
      * Prune all of the entries older than the given date.
      *
      * @param  \DateTimeInterface  $before
+     * @param  boolean  $skipException
      * @return int
      */
-    public function prune(DateTimeInterface $before)
+    public function prune(DateTimeInterface $before, $skipException)
```
driesvints, line 339: Same here

src/Contracts/PrunableRepository.php
gennarobotta: Should I make a PR to another branch or just wait until a new major release is schedule ?
I don't know if can help but the argument is optional.
Thanks anyway

lazer-hybiz:
It's already in the master... https://github.com/laravel/telescope/commit/7bfe65fe17905eb2fb4836b0346e09eb04374739

src/Contracts/PrunableRepository.php
driesvints: Doesn't matter if it's optional. You're changing the signature here which means any class implementing the contract will break. It's best to send this to master.

driesvints:
Seems this is indeed already on master. At this moment, we don't have a release date, sorry.

PR #1227, opened 2022-05-17
Add config key for environments telescope can be accessed in

hmennen90:
Currently, Laravel Telescope cannot be accessed in Environments like staging and production.

This PR adds a config key "environments" which contains a single value | array.

The AuthorizesRequests.php File reads this key or, if the key does not exists, it assumes only local access.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -20,6 +20,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Telescope Environments
+    |--------------------------------------------------------------------------
+    |
+    | These are the environments which can be set in .env where telescope can
+    | be accessed in. If accessing in environment which is not in this array
+    | you will get an HTTP 403 error.
+    |
+     */
+    'environments' => [
+        'local',
+        // e.g. 'staging', 'production'
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | Telescope Path
     |--------------------------------------------------------------------------
     |
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -28,6 +28,7 @@ return [
     | you will get an HTTP 403 error.
     |
      */
+
     'environments' => [
         'local',
         // e.g. 'staging', 'production'
--- a/src/AuthorizesRequests.php
+++ b/src/AuthorizesRequests.php
@@ -33,7 +33,7 @@ trait AuthorizesRequests
     public static function check($request)
     {
         return (static::$authUsing ?: function () {
-            return app()->environment('local');
+            return app()->environment(config('telescope.environments', ['local']));
         })($request);
     }
 }


driesvints:
> Currently, Laravel Telescope cannot be accessed in Environments like staging and production.

You need to configure: https://laravel.com/docs/9.x/telescope#dashboard-authorization

hmennen90:
@driesvints I already tried this so i took a look in the source files. We want to use Telescope on stage and production too.

The Gate is defined as following:
```

/**
     * Register the Telescope gate.
     *
     * This gate determines who can access Telescope in non-local environments.
     *
     * @return void
     */
    protected function gate()
    {
        Gate::define('viewTelescope', function ($user) {
            return in_array($user->email, [
                'hendrik@vomwege.gmbh',
                'henning@vomwege.gmbh',
                'info@vomwege.gmbh',
                'hendrik.mennen@mennen.online'
            ]);
        });
    }
```

driesvints:
That should enable it on production.

hmennen90:
But it doesn't because the AuthorizeRequests.php checks for local Environment only, not if the Gate allows.

If I change the AuthorizeRequests.php like in this PR, i have access to Telescope.

hmennen90:
@driesvints i just added a Test for the Accessibility.

using 'local' for environment check in AuthorizeRequests::check() Method, the Test fails caused by 403.

driesvints:
@hmennen90 It seems you're right but your solution isn't the correct one. The problem is that the call never reaches the custom gate set in your app's telescope service provider. `static::$authUsing` will always return `false` and I'm baffled that it doesn't work. I tried doing a deep dive but couldn't figure it out. Pinging our team to help look into this.

hmennen90:
@driesvints Part of the Problem seems to be, that the TelescopeServiceProvider is not called when the application boots.

I tried to manually register the TelescopeServiceProvider in config/app.php and then it worked

hmennen90:
It seems that auto-discovery does not work properly:

```
> Illuminate\Foundation\ComposerScripts::postAutoloadDump
> @php artisan package:discover --ansi
Discovered Package: dinhquochan/laravel-twig
Discovered Package: fruitcake/laravel-cors
Discovered Package: inertiajs/inertia-laravel
Discovered Package: intervention/image
Discovered Package: jenssegers/agent
Discovered Package: laravel/cashier
Discovered Package: laravel/fortify
Discovered Package: laravel/jetstream
Discovered Package: laravel/nova
Discovered Package: laravel/sail
Discovered Package: laravel/sanctum
Discovered Package: laravel/spark-stripe
Discovered Package: laravel/tinker
Discovered Package: laravel/ui
Discovered Package: mpociot/vat-calculator
Discovered Package: nesbot/carbon
Discovered Package: nunomaduro/collision
Discovered Package: pestphp/pest
Discovered Package: rap2hpoutre/fast-excel
Discovered Package: spatie/laravel-ignition
Discovered Package: tightenco/ziggy
Package manifest generated successfully.
105 packages you are using are looking for funding.

```

driesvints:
It does for me.

Screenshot 2022-05-17 at 13 10 37

And the TelescopeServiceProvider is part of my app.php

Screenshot 2022-05-17 at 13 11 06

hmennen90:
On a fresh Installation i can confirm this .... odd

taylorotwell:
All Laravel gate checks return `false` if you are not authenticated. If you are not in the local environment, Telescope will check the `viewTelescope` gate which, again, if you are not authenticated, will return false. Install Breeze and login to your application and then you will see it working and invoking the gate. I tested this on a fresh install.

hmennen90:
@taylorotwell I use Jetstream doing this :-) But i will try it out - thank you

PR #1248, opened 2022-09-05
Using custom table name, and adding schema for the table

aizuddin-yusoff:
User can choose to set the table to other table name, and also use schema for the table.
Default behavior for the table name is fixed to _**dbo**.telescope_entries_. 
In _Config > telescope.php_, just change the lines below: 
`'table_name' => [
        'telescope_entries'      => 'telescope_entries'
    ],` to 
`'table_name' => [
        'telescope_entries'      => 'schema.telescope_entries'
    ],`, the table name is now _**schema**.telescope_entries_
This enables multiple Telescope to be installed in the same database, provided each laravel use different schema.

The example above only for 1 table name, you can rename only 1 or all 3 table that are required.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -53,6 +53,20 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Telescope Custom Table Name
+    |--------------------------------------------------------------------------
+    | This configues custom name for table
+    |
+    */
+
+    'table_name' => [
+        'telescope_entries'      => 'telescope_entries',
+        'telescope_entries_tags' => 'telescope_entries_tags',
+        'telescope_monitoring'   => 'telescope_monitoring',
+    ],
+
+    /*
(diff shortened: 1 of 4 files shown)


driesvints:
I don't really see a reason in your PR description why this is needed?

aizuddin-yusoff:
@driesvints All of our Laravel projects need to be in saperate schema in the same database, and if telescope can only be installed in dbo, then only one instances of telescope can be installed.
At very least, people can also rename the table to suit their naming scheme better. As our project also renames this telescope to have sys_ prefix.

taylorotwell:

zakiaziz:
@aizuddin-yusoff hello thanks for the PR also. I'm wondering if you took @taylorotwell's advice and released this functionality as a package? I would also like this functionality on my end.

PR #1252, opened 2022-09-21
Add response tab with full details for Gate screen

ajira86:
I'm using `allow` and `deny`  Response in my Policies like this.

```
    public function viewAny(User $user): Response
    {
        if (...) {
            return $this->deny("User role is not authorized");
        }

        return $this->allow();
    }
```

This let me send a specific message when Authorization is refused but the data is ignored by telescope.

Here I propose to record and display the response content in a new tab next to `Arguments` one


taylorotwell:

Issue #1255, opened 2022-09-28
JobWatcher causes duplicate queries


mpskovvang:
- Telescope Version: 4.9.3
- Laravel Version: 9.29.0
- PHP Version: 8.0.21
- Database Driver & Version: PostgreSQL 9.6.6

### Description:

When `JobWatcher` is enabled, jobs with model properties and `SerializesModels` trait will cause duplicate queries.

The reason is that the job is unserialized twice - first by the queue worker and later by the `JobWatcher`. Every time the job is unserialized, the trait will restore every model property.

Debug backtraces:
```
{
"file": "/app/vendor/laravel/framework/src/Illuminate/Queue/CallQueuedHandler.php",
"line": 97,
"function": "unserialize",
...
}
```

```
{
"file": "/app/vendor/laravel/telescope/src/Watchers/JobWatcher.php",
"line": 216,
"function": "unserialize",
...
}
```

The `unserialize` happens when the `updateBatch` method tries to extract properties from the command.

https://github.com/laravel/telescope/blob/a8d769ea573f05a26656f7e20b3cf25daaac2e89/src/Watchers/JobWatcher.php#L186

The issue can be resolved by disabling recording. However, I doubt that is the preferable way to solve this since the queries remain but skips recording.

```PHP
Telescope::$shouldRecord = false;

$command = $this->getCommand($payload['data']);

Telescope::$shouldRecord = true;
```

### Steps To Reproduce:

Create a job:

```PHP
class Test implements ShouldQueue
{
    use SerializesModels;
    
    protected User $user;
    
    public function __construct(User $user)
    {
        $this->user = $user;
    }
    
    public function handle()
    {
        Log::debug($this->user->name);
    }
}
```

Dispatch job: `dispatch(new Test($user));`

Run queue worker: `php artisan queue:work`

driesvints:
I think ideally there indeed wouldn't be duplicate queries but maybe we should try the quick fix for now. Can you send in a PR?

driesvints:
I sent in a PR for this: https://github.com/laravel/telescope/pull/1257

PR #1257, opened 2022-09-30
[4.x] Do not record when unserializing command

driesvints:
This PR prevents Telescope from recording duplicate queries when unserializing a job. Ideally the duplicate queries shouldn't happen but at least this prevents duplicate date from being added to Telescope.

Fixes https://github.com/laravel/telescope/issues/1255

### DIFF
--- a/src/Watchers/JobWatcher.php
+++ b/src/Watchers/JobWatcher.php
@@ -183,7 +183,11 @@ class JobWatcher extends Watcher
      */
     protected function updateBatch($payload)
     {
+        Telescope::$shouldRecord = false;
+
         $command = $this->getCommand($payload['data']);
+
+        Telescope::$shouldRecord = true;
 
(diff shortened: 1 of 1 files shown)


mpskovvang:
Thanks!

Perhaps it should check the original state before re-enabling recording?

mpskovvang:
Some workaround ideas to completely avoid duplicate queries:
1. Use `preg_match` to extract the batch ID
2. Modify the `JobProcessed` event in laravel/framework to pass the unserialized data/payload (the job handler `$instance` property) using `getResolvedJob` method

driesvints:
> Perhaps it should check the original state before re-enabling recording?

Thanks, I sent in a PR for that: https://github.com/laravel/telescope/pull/1258

> Could we just do an instance check in the JobWatcher class with fallback to unserializing?

Sorry, I'm not sure if it's the right approach.

PR #1281, opened 2022-12-12
 adjust text-break layout with text-break class

thomas-goncalves-83:
adjust text-break layout with text-break class

In the case of very long text, the layout was breaking with this adjustment, it automatically fixes it, I increased it to 200 characters as well.


taylorotwell:
Hi there - can you provide before and after screenshots? Please mark as ready for review when done.

thomas-goncalves-83:
Antes
image

image

Depois
image
image

thomas-goncalves-83:
add the requested images

PR #1286, opened 2022-12-27
Add support for model keys cast to backed enums in FormatModel

sidquisaad:
As the title states, add support for model keys cast to backed enums in FormatModel. I have not scouted the whole codebase, there may be other places where a fix is needed to support backed enums for model keys.

I came across this bug while working on a project of mine. I had left this project alone for a couple of weeks before resuming it today. The issue didn't happen before, and I did a `composer update` when resuming, so this might be a  regression. I, unfortunately, do not have the time to check more thoroughly.

Hope this helps. Thank you.

### DIFF
--- a/src/FormatModel.php
+++ b/src/FormatModel.php
@@ -25,7 +25,7 @@ class FormatModel
             $keys = $model->getKey();
         }
 
-        if ($keys instanceof BackedEnum) {
+        if (interface_exists('BackedEnum') && ($keys instanceof BackedEnum)) {
             $keys = $keys->value;
         }
 
--- a/src/FormatModel.php
+++ b/src/FormatModel.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Telescope;
 
+use BackedEnum;
 use Illuminate\Database\Eloquent\Relations\Pivot;
 use Illuminate\Support\Arr;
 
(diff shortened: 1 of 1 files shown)


sidquisaad:
This will probably break support for PHP < 8.1. I'm not sure how compatibility is handled in this project, any help would be greatly appreciated.

driesvints:
We can't break apps on <8.1. 7.3 is the minimum version right now. You'll need to do a class_exists check I think.

sidquisaad:
Updated. The reasoning is, if the `interface_exists()` check fails it will short-circuit, and the rest of the `if` condition will not execute. Thus, this should maintain compatibility with PHP < 8.1.

sidquisaad:
I have just noticed also, it might be worthwhile to implement a similar check on the case at line 18. Let me know if that makes sense and I will refactor the whole thing to include all cases.

driesvints:
Line 18 seems fine to me.

sidquisaad:
I meant the block fulfilling the condition on line 18, specifically line 20 and 21. In that case, `$model->getAttribute` will also return the foreign/related key id cast to an enum, if such cast is set in the pivot instance. It makes sense given that `$model->getKey` on line 24 is just [a shortcut for `getAttribute`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Model.php#L1952).

In any case, I have provided the more complete solution that should take care of either case. LMK if you see any issues.

PR #1291, opened 2023-01-09
Fix for a redis cache bug

dgagn:
It fixes the redis cache bug.

https://github.com/laravel/telescope/issues/1023#issuecomment-962663627


taylorotwell:
Then no CLI information (queues, etc.) would be stored in Telescope.

PR #1301, opened 2023-02-07
Add customize table names

yangjisen:


driesvints:
I don't feel we should do this really. You've also not provided an explanation why you need this.

PR #1307, opened 2023-02-20
[4.x] Display Tags on Index Pages 

moisish:
Similar to the Horizon UI, I think it will be useful to display the tags on the index page for the watchers that support it.

The tags will only be displayed if they exists and on the following pages

- Requests
- Jobs
- Batches
- Cache
- Events
- Exceptions
- Gates
- Logs
- Mail
- Notifications
- Redis
- Views

As part of this PR, I have added a new model EntryTagModel and the relationship tags have been defined on the EntryModel. The `whereTag` method has been updated as part of this

image

image

image


taylorotwell:

moisish:
@taylorotwell will these changes not useful? The idea is to make it look similar to Horizon anyway

PR #1308, opened 2023-02-20
[4.x] Improvements for multiple Guards

moisish:
At the moment if you are using multiple guards on your app, there is not a way to identify which guard is being used on an entry.

The tag is always `Auth:1`.

I have updated the tag to include the full class name `{modelName}:{id}` as the screenshot bellow. 
In order to avoid any breaking changes to people that are monitoring tags at the moment for certain users I haven't removed the tag `Auth:{id}`

I have also added the guard name(s) on the Authenticated User Panel

image

### DIFF
--- a/resources/js/components/PreviewScreen.vue
+++ b/resources/js/components/PreviewScreen.vue
@@ -224,6 +224,13 @@
 
             <table class="table mb-0 card-bg-secondary table-borderless">
                 <tr>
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can you resubmit this with only the change to the user panel to show the guard name? The tag change is not needed as you can already add whatever tags you want: https://laravel.com/docs/10.x/telescope#tagging

PR #1312, opened 2023-02-21
[4.x] Allows to specify the log level

nunomaduro:
This merge request adds the `level` option to the `LogWatcher` in Telescope. The default value of the `level` option is debug, for backward compatibility. However, after merging this request, users will begin to use the updated configuration stub with error as the default log level.

It's important to note that regular deprecations are considered as warnings. Therefore, users will only see the deprecations in their telescope table if they set the log level to `warning` or lower.

One thing, up to debate, is that whether the `LOG_LEVEL` environment variable should be utilized in the config/telescope.php stub.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -149,7 +149,12 @@ return [
         ],
 
         Watchers\JobWatcher::class => env('TELESCOPE_JOB_WATCHER', true),
-        Watchers\LogWatcher::class => env('TELESCOPE_LOG_WATCHER', true),
+
+        Watchers\LogWatcher::class => [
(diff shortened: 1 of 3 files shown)


driesvints (approved):
Left one optional comment but other than that this lgtm.

hms5232:
> why is level default level "error" in stub? it was quite challenge to find why telescope is not showing info('debug string').

Agree! This should have a env and set default is `debug` to keep origin behaviour. Just like:

```php
'level' => env('TELESCOPE_LOG_WATCH_LEVEL', 'debug')
```

PR #1315, opened 2023-02-28
[4.x] Display Guard Name(s) on Auth Panel

moisish:
This is the updated PR of what Taylor requested on a previous PR here https://github.com/laravel/telescope/pull/1308 to display the guard name(s) on the Authenticated Panel.

### DIFF
--- a/src/IncomingEntry.php
+++ b/src/IncomingEntry.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Telescope;
 
+use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use Laravel\Telescope\Contracts\EntriesRepository;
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
What will be displayed for old entires that do not have the guard information stored?

moisish:
Will display N/A

image

taylorotwell:
Just hide the Guard label and value entirely if it is not present please. Please mark as ready for review again when done.

src/IncomingEntry.php
```diff
@@ -319,4 +320,24 @@ public function toArray()
...
+                return config("auth.providers.{$guard['provider']}.model");
+            })
+            ->filter(fn ($model) => get_class($this->user) === $model)
+            ->keys();
+
+        if ($guards->isNotEmpty()) {
+            return $guards->all();
+        }
+
+        return [config('auth.defaults.guard')];
```
taylorotwell, line 341: Eh, on second look I don't think is necessarily accurate. You could have multiple guards using the same model? That doesn't mean it's the guard that authenticated this request.

Issue #1318, opened 2023-03-09
Dump & Die with Telescope


finagin:
When you use the `dd` function when the Telescope is enabled, you will not get the result because the entry entries happens in Laravel terminate. We lose records of already completed database transactions, events, dumps, etc.
I see several ways to solve this situation and would like to discuss them before doing MR.

1. Change `DumpWatcher` so that when the `dd` function is called, a message is output, and not written to the database (but we still lose other records of this request).
2. Run Laravel `terminate` when `exit` function is called (but I don't think this is a good idea).
3. Register separate register_shutdown_function for Telescope

In my opinion, compatibility of 1 and 3 approaches is best for saving records and outputting `dd`.

PR #1320, opened 2023-03-09
[4.x] Record entries after dd or exit

finagin:
PR for issue #1318

## Problem
When you use the `dd` function when the Telescope is enabled, you will not get the result because the entry entries happens in Laravel terminate. We don't see `dd` result and lose records of already completed database transactions, events, dumps, etc.

## Possible solution
- Always show dump result when `dd` called
- Record entries by self termination handler after `dd` or `exit` if config option `telescope.self_termination` is `true`

### DIFF
--- a/src/Watchers/DumpWatcher.php
+++ b/src/Watchers/DumpWatcher.php
@@ -84,7 +84,7 @@ class DumpWatcher extends Watcher
 
     protected static function getCliDumper(string $basePath, string $compiledViewPath)
     {
-        $foundation = \Illuminate\Foundation\Console\CliDumper::class;
+        $foundation = 'Illuminate\Foundation\Console\CliDumper';
 
         return class_exists($foundation)
             ? new $foundation(new ConsoleOutput, $basePath, $compiledViewPath)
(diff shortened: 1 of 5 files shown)


src/Watchers/DumpWatcher.php
```diff
@@ -4,26 +4,26 @@
...
 
 class DumpWatcher extends Watcher
 {
     /**
      * The cache factory implementation.
-     *
-     * @var \Illuminate\Contracts\Cache\Factory
      */
-    protected $cache;
+    protected CacheFactory $cache;
```
driesvints, line 22: Please revert all type and docblock changes you did.

driesvints (changes requested):

src/Watchers/DumpWatcher.php
finagin: Done

taylorotwell:

finagin:
@taylorotwell, I’m feel absolutely certain that this code corrects a bug in the package. 

Now using the `dd` function looks like a failure: a white screen and no records in the Telescope. All because the recording occurs asynchronously in `$app->terminate`, but it does not reach it

PR #1328, opened 2023-03-19
Added color highlighting of duplicate database queries

andrey-helldar:
Now the list of requests shows only the number of duplicate requests, and in order to find which ones are duplicated, you have to manually compare them.

I suggest highlighting entries containing duplicate requests.

#### Before
image

#### After
image


taylorotwell:
I'm not sure that bright red color would clearly indicate duplicates. It looks like an error or something.

andrey-helldar:
@taylorotwell, I can offer the following options from the colors specified in the settings:

### Light mode
Grey:
image
Yellow:
image
Blue:
image

### Dark mode
Grey:
image
Yellow:
image
Blue:
image

Issue #1326, opened 2023-03-17
Highlighting duplicate queries from the database


andrey-helldar:
Now the list of requests shows only the number of duplicate requests, and in order to find which ones are duplicated, you have to manually compare them.

I suggest highlighting entries containing duplicate requests.

#### Current view
![image](https://user-images.githubusercontent.com/10347617/225940126-d6f14347-b2a7-4d13-b085-c0f83f0a1bad.png)

#### Suggest view
![image](https://user-images.githubusercontent.com/10347617/225939954-c0bd9c42-f134-44d5-936d-ca26abd49e44.png)

driesvints:
Feel free to attempt a PR 👍

driesvints:
@andrey-helldar I don't review prs myself, taylor does. Feel free to attempt whatever you prefer.

PR #1330, opened 2023-03-25
[4.x] Added color highlighting of duplicate database queries

andrey-helldar:
Now the list of requests shows only the number of duplicate requests, and in order to find which ones are duplicated, you have to manually compare them.

I suggest highlighting entries containing duplicate requests.

#### Before
image

#### After
image
image

For issue https://github.com/laravel/telescope/issues/1326


taylorotwell:
Sorry - still not in love with this. Will try to figure out something on my end.

nurdism:
@taylorotwell, any updates on this? This would be a great addition to the telescope, it's been over a year, and nothing has been added. I'm running into having a hard time trying to figure out where/what my duplicated queries are, especially when there are more than 10-20 queries in a request or job. I'd be willing to re-add/expand on this PR but I don't see a point if it just gets left with a "not in love with this" and not any actual feedback that is actionable.

Shadercloud:
Why is this not merged? Would be very useful.  For example find the duplicates here (it's not the insert jobs)....it's very difficult to find duplicates.

image

andrey-helldar:
Apparently either very few people use Telescope, or no one looks at duplicate requests :)

Issue #577, opened 2019-03-05
Some jobs stay in Pending Status. More info given to help re-create


jonnywilliamson:
@themsaid 

So I have seen the previous issues related to this 
https://github.com/laravel/telescope/issues/507
https://github.com/laravel/telescope/issues/503
https://github.com/laravel/telescope/issues/411

And I'm suffering the same problem. *Some* of my jobs are reporting as PENDING even though I know they have completed successfully, and also HORIZON shows the same job as being done correctly.

I've spent a lot of time trying to recreate this using as minimal amount of code as possible, and I hope I've succeeded in providing you with an example that will replicate on your system.

Very quickly as an overall picture. 

- I have a LONG running job that gets dispatched.
- Inside that job other events (jobs) get dispatched too, in this case its a broadcast event to update user with some info.
- The LONG running job will always show correct status.
- The broadcast events/jobs do NOT show correct status on TELESCOPE, but DO on horizon.
- If queues are NOT used (ie sync) everything is OK.
- A different queue is used between Long runing job and Broadcast jobs.

Here's some code.

```php
// web.php
Route::get('/test', function () {
    dispatch(new \App\Jobs\LongJob());
    return "done";
});

```

```php
// \App\Jobs\LongJob.php
<?php

namespace App\Jobs;

use App\Events\SendUpdate;
use Illuminate\Bus\Queueable;
use Illuminate\Queue\SerializesModels;
use Illuminate\Queue\InteractsWithQueue;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Foundation\Bus\Dispatchable;

class LongJob implements ShouldQueue
{
    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    public $tries = 1;
    public $timeout = 120;

    public function handle()
    {
        sleep(5);
        event(new SendUpdate("The first time is:" . time()));
        sleep(5);
        event(new SendUpdate("The second time is:" . time()));
        sleep(10);
    }
}
```

```php
// App\Events\SendUpdate.php
<?php

namespace App\Events;

use Illuminate\Broadcasting\Channel;
use Illuminate\Queue\SerializesModels;
use Illuminate\Broadcasting\PrivateChannel;
use Illuminate\Foundation\Events\Dispatchable;
use Illuminate\Broadcasting\InteractsWithSockets;
use Illuminate\Contracts\Broadcasting\ShouldBroadcast;

class SendUpdate implements ShouldBroadcast
{
[...]

PR #1336, opened 2023-04-17
Retry updating data that cannot be queried in JobWatcher

miaotiao:
Fixed bug [#577](https://github.com/laravel/telescope/issues/577)

And  I think the [reason](https://github.com/laravel/telescope/issues/577#issuecomment-1510645312)  for this bug is that before updating the data, the data does not exist (The request has not ended and the data has not been inserted yet. See: [file ListensForStorageOpportunities@storeEntriesBeforeTermination](https://github.com/laravel/telescope/blob/4.x/src/ListensForStorageOpportunities.php) )

### DIFF
--- a/src/Storage/DatabaseEntriesRepository.php
+++ b/src/Storage/DatabaseEntriesRepository.php
@@ -208,6 +208,8 @@ class DatabaseEntriesRepository implements Contract, ClearableRepository, Prunab
      */
     public function update(Collection $updates)
     {
+        $failUpdates = [];
+
         foreach ($updates as $update) {
             $entry = $this->table('telescope_entries')
                             ->where('uuid', $update->uuid)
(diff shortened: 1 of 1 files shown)


taylorotwell:
This greatly complicates Telescope because now you have to run a queue worker locally to make this work properly. 

You are correct the reason for the bug is the job finishes processing before original request that queued it even finishes. But, not sure this is the correct solution.

miaotiao:
> Fixed bug [#577](https://github.com/laravel/telescope/issues/577)
> 
> And I think the [reason](https://github.com/laravel/telescope/issues/577#issuecomment-1510645312) for this bug is that before updating the data, the data does not exist (The request has not ended and the data has not been inserted yet. See: [file ListensForStorageOpportunities@storeEntriesBeforeTermination](https://github.com/laravel/telescope/blob/4.x/src/ListensForStorageOpportunities.php) )



> This greatly complicates Telescope because now you have to run a queue worker locally to make this work properly.
> 
> You are correct the reason for the bug is the job finishes processing before original request that queued it even finishes. But, not sure this is the correct solution.

Ok. I will try again with a correct solution.

PR #1337, opened 2023-04-18
Fix issue with empty Telescope/Horizon path disables recording

### LINKED ISSUE #1300: Empty (but not null) TELESCOPE_PATH and/or HORIZON_PATH disables recording.
- Telescope Version: 4.13.0
- Laravel Version: 9.5.2
- PHP Version: 8.2.1
- Database Driver & Version: mysql 8.0.31

### Description:
An empty (but not `NULL`) value in _either_ `config('telescope.path')` or `config('horizon.path')` incorrectly disables all Telescope recording. The lines [here](https://github.com/laravel/telescope/blob/4.x/src/Telescope.php#L235) and [here](https://github.com/laravel/telescope/blob/4.x/src/Telescope.php#L239-L241) add the plain wildcard `'*'` to the collection of unapproved URIs when either config value is defined as an empty string.

#### Use Case
I want to host Telescope and Horizon at `telescope.example.com/` and `horizon.example.com/`, respectively. The explicit `null` checks in [Telescope::requestIsToApprovedUri](https://github.com/laravel/telescope/blob/4.x/src/Telescope.php#L246-L267) make this impossible, even though the configuration supports the dashboard(s) loading with those settings.

### Steps To Reproduce:
1. Install Telescope.
2. Create file `config/horizon.php` with the content: `<?php return ['path' => '']; `
3. Notice no recording happens at any application URIs.

syahid246:
Fixes #1300.

This pull request fixes issue #1300. The issue caused recording to be disabled when the Telescope or Horizon path was set to an empty string `''`. With this fix, it's now possible to host Telescope and Horizon at different subdomains, without disabling recording.

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -231,11 +231,13 @@ class Telescope
             collect([
                 'telescope-api*',
                 'vendor/telescope*',
-                (config('horizon.path') ?? 'horizon').'*',
                 'vendor/horizon*',
             ])
             ->merge(config('telescope.ignore_paths', []))
-            ->unless(is_null(config('telescope.path')), function ($paths) {
+            ->unless(blank(config('horizon.path')), function ($paths) {
+                return $paths->prepend(config('horizon.path').'*');
+            })
+            ->unless(blank(config('telescope.path')), function ($paths) {
                 return $paths->prepend(config('telescope.path').'*');
             })
             ->all()


taylorotwell:

syahid246:
@taylorotwell

> If possible, please consider releasing your code as a package so that the community can still take advantage of your contributions!

I can extends the `Telescope` class and override `requestIsToApprovedUri` method but it may useful for other developers that need to host telescope at root of different subdomain.


> If you feel absolutely certain that this code corrects a bug in the framework, please "@" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

I want to host telescope via `http://telescope.example.local/` instead of `http://telescope.example.local/telescope` so I use this .env config:
```env
TELESCOPE_DOMAIN=telescope.example.local
TELESCOPE_PATH=
```
The problem I wanna solve is telescope doesn't recording anything with that .env config.

The lines [here](https://github.com/laravel/telescope/blob/4.x/src/Telescope.php#L234) and [here](https://github.com/laravel/telescope/blob/4.x/src/Telescope.php#L238-L240) add the plain wildcard `'*'` to the collection of unapproved URIs, when

syahid246:
@taylorotwell

> If possible, please consider releasing your code as a package so that the community can still take advantage of your contributions!

I understand your suggestion, however, I think the fix that I provided can be useful for other developers who need to host Telescope and/or Horizon at the root of different subdomains. They may not want to install a separate package for this.


> If you feel absolutely certain that this code corrects a bug in the framework, please "@" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

I want to host telescope via `http://telescope.example.local/` instead of `http://telescope.example.local/telescope` so I tried this .env config:
```env
TELESCOPE_DOMAIN=telescope.example.local
TELESCOPE_PATH=null
```
It does recording requests, but Telescope just show `Scanning...`, so I try set the `TELESCOPE_PATH` to an empty string
```env
TELESCOPE_PATH=
```
But the lines [here](https://github.com/laravel/telescope/blob/4.x/src/Telescope.php#L234) and [here](https://github.com/laravel/telescope/blob/4.x/src/Telescope.php#L238-L240) add the plain wildcard `'*'` to the collection of unapproved URIs, when either `config('telescope.path')` or `config('horizon.path')` is defined as an empty string (**all request disapproved for recording**).

Please let me know if you need any further clarification. Thank you.

PR #1341, opened 2023-04-23
Excluding SQL Statements in QueryWatcher for Improved Productivity

Speedfighter:
Since I started using Telescope, I noticed a missing feature to exclude SQL statements. I only want to see changes to my database in my productive area (INSERT/CREATE/DELETE), so I filtered out all SELECT statements. It is easy to select multiple SQL keywords at the beginning of an SQL statement and exclude them from showing/storing in the QueryWatcher.


taylorotwell:
Hey there - you can build custom logic to ignore any incoming Telescope entry:

https://laravel.com/docs/10.x/telescope#filtering

PR #1346, opened 2023-05-05
[4.x] fix: json_encode gives 502 Bad Gateway in EventWatcher

n1crack:
When an event payload has a JsonSerializable class, there is no need to decode and encode the object. It will fail by throwing 502 errors if the object has something like the service container as a property.

In my case, this fix solves the problem.


taylorotwell:

PR #1347, opened 2023-05-18
Update to use Jenssegers\Mongodb\Eloquent\Model

GodfreysDragon:
Commit 2: 'Resets the array keys to consecutive integers so the insertion works on non-MySQL inserts.'
Commit 3: 'Added ability to configure default sorting column.'


GodfreysDragon:
Wrong repo, whoops.

PR #1348, opened 2023-05-19
[4.x] Add duration to HTTP Client watcher

DougSisk:
This adds logging duration via the HTTP Client response's `GuzzleHttp\TransferStats`. `getTransferTime()` returns in seconds, so I'm converting to milliseconds to match the request design.


taylorotwell:
Thanks

Issue #1331, opened 2023-03-29
Job Status Pending


antoniovalenzuela:
- Telescope Version: 4.14
- Laravel Version: 10.4
- PHP Version: 8.1
- Database Driver & Version: MariaDB

### Description:
Same issues 1111 and 1113

I use Laravel with Redis and Horizon, with service supervisor.

Queues are saved and processed correctly, but in Telescope maintains the "pending" state

I did two tests, one queue pending and one processed. View image.

Attached diagram.

![Screenshot_2](https://user-images.githubusercontent.com/28113584/228656760-ad9500d2-fcbc-49d4-98aa-52896c5d6af7.jpg)
![Screenshot_3](https://user-images.githubusercontent.com/28113584/228656768-7d69066e-6adb-47ac-b452-d70355323531.jpg)

antoniovalenzuela:
I found the solution.

After installing Telescope it is necessary to restart processing of the queues.

/etc/init.d/supervisor restart

![Screenshot_4](https://user-images.githubusercontent.com/28113584/228658467-b680fbb0-34b0-4d2d-8027-6c6327facc41.jpg)

Issue #1113, opened 2021-08-12
Wrong "pending" Jobs status when using Laravel Horizon


ilyavaiser:
Telescope Version: v4.6.1
Laravel Version: v8.48.2
PHP Version: 7.4.22
Database Driver & Version:
Description:
Telescope shows pending status for jobs when using Horizon, even if the task is successful.

I am using Queued Laravel Notifications and they work great. The Horizon Dashboard displays the jobs information correctly. Notifications are sent to users. Everything is fine, but when I look at the list of jobs in Telescope, they are all displayed in the pending status.

I looked in Redis, there are no these jobs in pending status, the queue is empty.

Steps To Reproduce:
Create any Queues Notification Job and process it through Horizon

-----

https://github.com/laravel/telescope/issues/1111
https://github.com/laravel/telescope/issues/577
https://github.com/laravel/telescope/issues/507
https://github.com/laravel/telescope/issues/503
https://github.com/laravel/telescope/issues/411

Yes, I read the previous issues related to this problem and discussions on the support sites. Nothing helped.

driesvints:
Heya, thanks for reporting.

We'll need more info and/or code to debug this further. Can you please create a repository with the command below, commit the code that reproduces the issue as separate commits on the main/master branch and share the repository here? Please make sure that you have the [latest version of the Laravel installer](https://github.com/laravel/installer) in order to run this command. Please also make sure you have both Git & [the GitHub CLI tool](https://cli.github.com) properly set up.

```bash
laravel new bug-report --github="--public"
```

Please do not amend and create a separate commit with your custom changes. After you've posted the repository, we'll try to reproduce the issue.

Thanks!

driesvints:
Closing this issue because it's inactive, already solved, old or not relevant anymore. Feel to open up a new issue if you're still experiencing this.

antoniovalenzuela:
https://github.com/laravel/telescope/issues/1331

Issue #690, opened 2019-08-16
[Bug Report] failed and processed jobs are not recorded if there was no pending job recorded


christophmayrhofer:
- Telescope Version: 2.0
- Laravel Version: 5.8
- PHP Version: 7.2
- Database Driver & Version: mysql

### Description:
As far as I can tell, failed and processed job entries are treated as updates to an existing job entry (```DatabaseEntriesRepository->update()```). This assumes that there is already an entry in the DB (that should be created when the job is dispatched).

However, there may not be a an entry for the pending job in the DB (e.g. if telescope was disabled when the job was dispatched). A solution could be to create a new job entry if there is no existing pending job found.

I also never get an IncomingEntry passed to ``` Telescope::filterBatch ``` for failed jobs. Therefore the ```$entry->isFailedJob()``` filter will never return true.

### Steps To Reproduce:
Disable telescope
Dispatch job
Enable telescope
Process job (php artisan queue:work)
--> The job (regardless if failed or processed) does not show up in telescope.

taylorotwell:
We likely won't fix this as it would be a major change to how Telescope works. It is simply a limitation that if Telescope is not enabled when the job is dispatched it will not show up.

Issue #543, opened 2019-01-26
Telescope Job stays on pending status with SQS


llioor:
Queue Driver: SQS
Working on laradock (docker container).
The commend I run: `php /var/www/artisan queue:work sqs --sleep=3 --tries=3`

When telescope is on "Play" status, the Job stays on pending status and never processed. 
I see in the worker log 3 "processing.." rows:
```
[2019-01-26 23:18:46][1fbee0b4-3b81-4688-85c5-5a698ab25f01] Processing: App\Jobs\Client\SendVerificationCode
[2019-01-26 23:18:47][1fbee0b4-3b81-4688-85c5-5a698ab25f01] Processing: App\Jobs\Client\SendVerificationCode
[2019-01-26 23:18:50][1fbee0b4-3b81-4688-85c5-5a698ab25f01] Processing: App\Jobs\Client\SendVerificationCode
```

When telescope is paused (Play/Pause button), the Job is Processed:
```
[2019-01-26 23:19:49][d07c1bff-9410-499e-bdb8-a3e9ba09969e] Processing: App\Jobs\Client\SendVerificationCode
[2019-01-26 23:19:53][d07c1bff-9410-499e-bdb8-a3e9ba09969e] Processed:  App\Jobs\Client\SendVerificationCode
```

Thanks

themsaid:
I can't replicate this.

llioor:
@themsaid  can I share more info? how telescope can influence on the queue?

themsaid:
Well since I can't replicate it and it hasn't been reported by other people so maybe it's something in your app.

I'm going to close this, but please ping me in case someone else could replicate the issue or gather more details on the reason why this might be happening.

medvesekg:
Hi,
I had the same problem while using laradock: if telescope was enabled jobs would stay pending and never get processed.

In my case the problem was that the bcmath extension was not installed in the php-worker container. I resolved this by adding `RUN docker-php-ext-install bcmath` to the container's dockerfile

fmsedrez:
Hi,
@themsaid I had the same issue as @llioor, but with the comment of @medvesekg I change the Laradock enviroment configuration adding PHP_WORKER_INSTALL_BCMATH=true.

PR #1349, opened 2023-05-24
[4.x] Fixes jobs saying "pending" forever

nunomaduro:
This pull request fixes the "pending" job status (https://github.com/laravel/telescope/issues/1331, https://github.com/laravel/telescope/issues/1113, https://github.com/laravel/telescope/issues/577, https://github.com/laravel/telescope/issues/507, https://github.com/laravel/telescope/issues/543, https://github.com/laravel/telescope/issues/690, https://github.com/laravel/telescope/issues/503, https://github.com/laravel/telescope/issues/411), by simply delaying the "failed" updates to performed a few seconds after via queued job.

### DIFF
--- a/src/Contracts/EntriesRepository.php
+++ b/src/Contracts/EntriesRepository.php
@@ -37,7 +37,7 @@ interface EntriesRepository
      * Store the given entry updates, and return the failed updates.
      *
      * @param  \Illuminate\Support\Collection|\Laravel\Telescope\EntryUpdate[]  $updates
-     * @return \Illuminate\Support\Collection
+     * @return \Illuminate\Support\Collection|null
      */
     public function update(Collection $updates);
 
--- a/src/Contracts/EntriesRepository.php
+++ b/src/Contracts/EntriesRepository.php
@@ -34,7 +34,7 @@ interface EntriesRepository
     public function store(Collection $entries);
 
     /**
-     * Store the given entry updates, and return the failed updates.
+     * Store the given entry updates and return the failed updates.
      *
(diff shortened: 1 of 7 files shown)


malohtie:
@nunomaduro  is it normal to see a lot of this queue jobs from this class ? 
image

nunomaduro:
@malohtie Yes, I am assuming you have an even bigger number of 'other' queued jobs, right?

malohtie:
@nunomaduro yes a lot 

image

nunomaduro:
So that number of "ProcessPendingUpdates" is expected, as some of your queued job jobs are being finished before even you insert them in telescope.

peterpan666:
I'm seeing these jobs dispatched to the default queue too. What does it do ? In what case is it dispatched ? I can't find any documentation about it.

peterpan666:
Hi and thanks for your reply but I don't use Horizon.

To me the problem is that an undocumented job is dispatched on an unconfigurable queue, to do I don't know what or when.

I'm not sure how should I handle it at this time and why should I handle it.

lucaspanik:
@nunomaduro 

Could this PR be a possible solution to this problem on the horizon?
https://github.com/laravel/horizon/issues/1034

mikedodd:
This is making millions of jobs for me, my default queue has gone from 2 threads to 30, just to keep up.

all jobs have specific queues in Horizon and so this is all to handle this patch.

I can ignore them from the completed queue in horizon but it will still be running 30 threads just for this job. Is there any way to disable this as its causing my servers to have a mini melt down :)

PR #1350, opened 2023-05-26
Prevent Memory Leak when using hideRequestParameters and hideRequestHeaders

danilopinotti:
This PR aims to prevent a memory leak when we call the method `hideRequestHeaders` or/and `hideRequestParameters` multiple times with the same parameters. Ex:

```php
Telescope::hideRequestHeaders(['cookie', 'x-csrf-token']);
Telescope::hideRequestHeaders(['cookie', 'x-csrf-token']);
```

It seems like an inoffensive issue, but it causes a memory leak in tests:
```php
class MemoryLeakTest extends TestCaseUnit
{
    public function test_leaks_memory_on_1000_iterations()
    {
        $this->app->flush();
        $this->app = null;

        for ($i = 1; $i < 500; ++$i) {
            $this->createApplication()
                ->flush();

            if (! ($i % 10)) {
                dump('Using ' . ((int) (memory_get_usage(true) / (1024 * 1024))) . 'MB as ' . $i . ' iterations.');
            }
        }

        $this->app = $this->createApplication();
    }
}
```
> The code above was copied from https://darkghosthunter.medium.com/laravel-fixing-memory-leaks-on-tests-4fe87b87f0ad

### DIFF
--- a/src/Telescope.php
+++ b/src/Telescope.php
@@ -726,10 +726,10 @@ class Telescope
      */
     public static function hideRequestHeaders(array $headers)
     {
-        static::$hiddenRequestHeaders = array_unique(array_merge(
+        static::$hiddenRequestHeaders = array_values(array_unique(array_merge(
             static::$hiddenRequestHeaders,
             $headers
-        ));
+        )));
 
         return new static;
     }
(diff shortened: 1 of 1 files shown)


driesvints:
This is definitely good but just fyi: Telescope shouldn't be enabled in tests.

danilopinotti:
> This is definitely good but just fyi: Telescope shouldn't be enabled in tests.

But even with the option `TELESCOPE_ENABLED=false`, the provider is initialized (except when it be configured as this session: https://laravel.com/docs/10.x/telescope#local-only-installation)

Issue #1353, opened 2023-06-02
Serialization of 'Closure' is not allowed


andrewminion-luminfire:
### Telescope Version

4.14.4

### Laravel Version

9.52.8

### PHP Version

8.1.11

### Description

As a result of https://github.com/laravel/telescope/pull/1349, I’m seeing `Exception: Serialization of 'Closure' is not allowed` errors in my app.

Context:

- The app uses redis queue workers and Horizon
- The exception occurs while the queue worker is processing a job

It seems that it’s due to the anonymous `fn()` [here](https://github.com/laravel/telescope/commit/9bbfc8f32fc26e2af85e1130ecde6a85859efa4d#diff-82c75df4b09d9a074291ebb138286af7e36da2156e8f2a9077218fc5f2e2dd97R671-R673), and it ultimately fails [here](https://github.com/laravel/framework/blob/c60191f33a88f91685634b157b4dc515381fccd1/src/Illuminate/Queue/Queue.php#L158)

### Steps To Reproduce

I haven’t yet nailed down exactly which job causes this

andrewminion-luminfire:
It’s happening when a job is [processed](https://github.com/laravel/telescope/blob/08930adb80b77f67d561ac1b9bb5f6a7e80d8791/src/ListensForStorageOpportunities.php#L85-L87), [fails](https://github.com/laravel/telescope/blob/08930adb80b77f67d561ac1b9bb5f6a7e80d8791/src/ListensForStorageOpportunities.php#L89-L91), and when the app is [terminating](https://github.com/laravel/telescope/blob/08930adb80b77f67d561ac1b9bb5f6a7e80d8791/src/ListensForStorageOpportunities.php#L64-L66)

andrewminion-luminfire:
@nunomaduro here you go; there are actually four distinct stack traces:

<details>
<summary>stacktrace 1</summary>

```
Exception Serialization of 'Closure' is not allowed
(unknown method)(/home/staging/example.com/vendor/laravel/framework/src/Illuminate/Queue/Queue.php:158)
serialize(/home/staging/example.com/vendor/laravel/framework/src/Illuminate/Queue/Queue.php:158)
Illuminate\Queue\Queue::createObjectPayload(/home/staging/example.com/vendor/laravel/framework/src/Illuminate/Queue/Queue.php:127)
[...]

PR #1354, opened 2023-06-02
[4.x] Avoids serialialisation of exception traces

nunomaduro:
This pull request addresses the problem described in issue https://github.com/laravel/telescope/issues/1353. Currently, when a job fails, we attempt to queue an "update" for the failed job status through ProcessPendingUpdates even before it exists in Telescope. However, there is an issue with this approach. The update may include a stack trace that contains references to closures, which, as we are aware, cannot be serialized.

Note that, the arguments are not even used on the Telescope UI, so we can remove them safely.

### DIFF
--- a/tests/Watchers/NotificationWatcherTest.php
+++ b/tests/Watchers/NotificationWatcherTest.php
@@ -44,7 +44,7 @@ class NotificationWatcherTest extends FeatureTestCase
         $this->assertFalse($entry->content['queued']);
         $this->assertStringContainsString(is_array($route) ? implode(',', $route) : $route, $entry->content['notifiable']);
         $this->assertSame($channel, $entry->content['channel']);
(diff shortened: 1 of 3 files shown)


tests/Watchers/NotificationWatcherTest.php
```diff
@@ -44,7 +44,7 @@ private function performNotificationAssertions($channel, $route)
         $this->assertFalse($entry->content['queued']);
         $this->assertStringContainsString(is_array($route) ? implode(',', $route) : $route, $entry->content['notifiable']);
         $this->assertSame($channel, $entry->content['channel']);
-        $this->assertNull($entry->content['response']);
+        $this->assertEmpty($entry->content['response']);
```
nunomaduro, line 47: Due: https://github.com/laravel/framework/pull/47310.

PR #1355, opened 2023-06-02
[4.x] Show list of headers when present

timacdonald:
When an array of headers is present, we now show the array rather than only the first value.

## Incoming Application Request

Screen Shot 2023-06-03 at 11 57 03 am

## Outgoing HTTP Client Request

Request see "x-foo"

Screen Shot 2023-06-03 at 11 32 43 am

Response see "set-cookie"

Screen Shot 2023-06-03 at 11 53 28 am

Went wit comma separated list instead of an actual array, as it creates consistency between those that already use the comma as a list mechanism and those that do not.


timacdonald:
Converted to draft as this is an issue in a few more places that I will also address.

PR #1364, opened 2023-07-24
[4.x] Add hotkey to clear entries without confirmation: CMD+K 

bradleybernard:
### Reasoning
`Command + K` is a common "clear" hotkey, that is supported in various terminals, Ray, and other dev tools. My telescope log quickly fills up, and when I want to test my workflow from start to finish, I often reach for clear entries button and click yes to confirm. It would be great to have that hotkey'd so I could go to my terminal, Ray, and Telescope to get a clean slate across all dev/debug tools, to reduce cognitive load and improve DevEx. 

### Logic
Super simple, add keydown event listener and if it matches our desired keystroke combination, call existing `clearEntries` function but pass in a flag to not require confirmation!

### Demo video
https://github.com/laravel/telescope/assets/1431995/e16d88e5-2e15-4c14-9668-dfae5bcc6c8c


taylorotwell:
Thanks

bradleybernard:
Sweeeet, thanks Taylor! 👌

driesvints:
Isn't CMD+K more used with search boxes? For almost every docs side, including Laravel's, I get a search box when I use it. In this case I'd expect it to focus on the search input.

bradleybernard:
For docs it is CMD+K, but for any other dev tools like terminals or things with scroll back history, CMD+K is analogous to clear / hide. 

For Telescope, I imagine the main use case would be clear, but we can always shift it to be something else if you want to pin that to search. My guess is clear is used much more often than search, but I have no data to back that up 😅

Lemme know! IMO clear is very intuitive and the shortcuts for tools are context dependent, I see Telescope as a primary debug tool, not search being the main piece. Open to other thoughts tho 😇

PR #1367, opened 2023-07-31
Enhancement: Multi-Tag Filtering and Search Functionality

ObaydaAlesawi:
### **Description:**

This Pull Request proposes the introduction of a significant enhancement to the existing search functionality in the Telescope. Specifically, it enables users to filter and search using multiple tags simultaneously for requests.

### **Why is this change necessary?**

The current tag-based search and filter system only allows one tag per search. With the new multi-tag search and filtering, users can have more granular control over their search results, allowing for a more thorough and precise search experience.

### DIFF
--- a/src/Storage/EntryModel.php
+++ b/src/Storage/EntryModel.php
@@ -116,8 +116,17 @@ class EntryModel extends Model
     protected function whereTag($query, EntryQueryOptions $options)
     {
         $query->when($options->tag, function ($query, $tag) {
-            return $query->whereIn('uuid', function ($query) use ($tag) {
-                $query->select('entry_uuid')->from('telescope_entries_tags')->whereTag($tag);
+            $tags = explode(',', $tag);
+            if (empty($tags)) {
+                return $this;
+            }
+            return $query->whereIn('uuid', function ($query) use ($tags) {
+                $query->select('entry_uuid')->from('telescope_entries_tags');
+                foreach ($tags as $tag) {
+                    $query->whereIn('entry_uuid', function ($query) use ($tag) {
+                        $query->select('entry_uuid')->from('telescope_entries_tags')->where('tag', trim($tag));
+                    });
+                }
             });
         });
 
--- a/src/Storage/EntryModel.php
+++ b/src/Storage/EntryModel.php
@@ -120,6 +120,7 @@ class EntryModel extends Model
             if (empty($tags)) {
                 return $this;
             }
+
             return $query->whereIn('uuid', function ($query) use ($tags) {
                 $query->select('entry_uuid')->from('telescope_entries_tags');
                 foreach ($tags as $tag) {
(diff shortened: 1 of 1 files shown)


src/Storage/EntryModel.php
```diff
@@ -116,8 +116,18 @@ protected function whereBatchId($query, EntryQueryOptions $options)
     protected function whereTag($query, EntryQueryOptions $options)
     {
         $query->when($options->tag, function ($query, $tag) {
-            return $query->whereIn('uuid', function ($query) use ($tag) {
-                $query->select('entry_uuid')->from('telescope_entries_tags')->whereTag($tag);
+            $tags = explode(',', $tag);
+            if (empty($tags)) {
+                return $this;
```
crynobone, line 121:
```suggestion
                return $query;
```

src/Storage/EntryModel.php
```diff
@@ -116,8 +116,18 @@ protected function whereBatchId($query, EntryQueryOptions $options)
     protected function whereTag($query, EntryQueryOptions $options)
     {
         $query->when($options->tag, function ($query, $tag) {
-            return $query->whereIn('uuid', function ($query) use ($tag) {
-                $query->select('entry_uuid')->from('telescope_entries_tags')->whereTag($tag);
+            $tags = explode(',', $tag);
+            if (empty($tags)) {
```
crynobone, line 120:
```suggestion
            $tags = collect(explode(',', $tag))->map(fn ($tag) => trim($tag));

            if ($tags->isEmpty()) {
```

crynobone:
Overall this looks good, but do have custom whether someone might already using `,` in custom tags since Telescope doesn't sanitize the value at the moment.

Algeneralo:
> Overall this looks good, but do have concern whether someone might already using `,` in custom tags since Telescope doesn't sanitize the value at the moment.

I'm not convinced that's the standard way to insert a tag, but how about this approach: 
If a tag contains a comma `,` the user should format their search like this - 
`"tag1,data", tag2, tag3`.
 What are your thoughts on this? @crynobone

taylorotwell:
I would be ok with that @Algeneralo 

We can also just let it roll like this for now.

faustoFF:
After making this change, I found that filtering by tag is extremely slow on a "large" database (I have a ~20GB separate Telescope database in MariaDB). The Telescope web interface hits a 504 error (after 1 minute of waiting) when filtering. I found that the database takes a long time to execute this query:

```sql
select 
  * 
from 
  `telescope_entries` 
where 
  `type` = ? 
  and `uuid` in (
    select 
      `entry_uuid` 
    from 
      `telescope_entries_tags` 
    where 
      `entry_uuid` in (
        select 
          `entry_uuid` 
        from 
          `telescope_entries_tags` 
        where 
          `tag` in (?)
      )
  ) 
order by 
  `sequence` desc 
limit 
  50
```

I believe the second `IN` clause contains a redundant subquery, and the resulting query should be optimized and simplified to:

```sql
select 
  * 
from 
  `telescope_entries` 
where 
  `type` = ? 
  and `uuid` in (
    select 
      `entry_uuid` 
    from 
      `telescope_entries_tags` 
    where 
      `tag` in (?)
  ) 
order by 
  `sequence` desc 
limit 
  50
```

PR #1378, opened 2023-08-24
Test Improvements

crynobone:
also exclude all other packages.


driesvints:
Does it still make sense to add Ray if we exclude everything?

crynobone:
Ray and Laravel Dusk has extra conditions and not covered by `*`.

driesvints:
Thanks Mior!

PR #1385, opened 2023-09-22
Fix duplicated queries on multiple connections

February30th:
Fixes duplicated query count when running same sql statement on multiple connections.

### DIFF
--- a/resources/js/components/RelatedEntries.vue
+++ b/resources/js/components/RelatedEntries.vue
@@ -144,7 +144,7 @@
             queriesSummary() {
                 return {
                     time: _.reduce(this.queries, (time, q) => { return time + parseFloat(q.content.time) }, 0.00).toFixed(2),
-                    duplicated: this.queries.length - _.size(_.groupBy(this.queries, (q) => { return q.content.hash })),
+                    duplicated: this.queries.length - _.size(_.groupBy(this.queries, (q) => { return `${q.content.hash}-${q.content.connection}` })),
                 };
             },
 


February30th:
Fetching activity log of multiple databases:

```
SELECT * FROM activity_log; // connection: db-1
SELECT * FROM activity_log; // connection: db-2
```

The queries get flagged as duplicated, despite running in two different databases, which is not accurate.

crynobone:
How does it handle the following?

```sql
SELECT * FROM activity_log; // connection: db-1
SELECT * FROM activity_log; // connection: db-2
SELECT * FROM activity_log; // connection: db-1
SELECT * FROM activity_log; // connection: db-2
SELECT * FROM activity_log; // connection: db-1
```

zhoubiaobiao:
ignore_paths not work

return ! $request->is(
            collect([
                'telescope-api*',
                'vendor/telescope*',
                'horizon*',
                'vendor/horizon*',
            ])
                ->merge(config('telescope.ignore_paths', []))
                ->unless(is_null(config('telescope.path')), function ($paths) {
                    return $paths->prepend(config('telescope.path').'*');
                })
                ->all()
        );

change:

$ignore = collect([
            'telescope-api*',
            'vendor/telescope*',
            'horizon*',
            'vendor/horizon*',
        ])
            ->merge(config('telescope.ignore_paths', []))
            ->unless(is_null(config('telescope.path')), function ($paths) {
                return $paths->prepend(config('telescope.path').'*');
            })
            ->all();
        foreach ($ignore as $path) {
            if($request->is($path)) return true;
        }

February30th:
5 duplicated queries, should be 3 duplicated queries (2 on connection db-1) and 1 duplicated query on db-2

crynobone:
Wouldn't that make the change above invalid?

```
queries.length - size(...duplicates)

5 - size([2, 3]) = 3
```

crynobone:
> 5 total queries, should be 3 duplicated queries (2 on connection db-1) and 1 duplicated query on db-2

This make sense.

crynobone (approved):

PR #1389, opened 2023-10-09
Call shouldIgnore method in every watchers

Pythagus:
Today, the package sometimes defines the shouldIgnore method as protected, sometimes as private. Sometimes the method is even not defined. This is sad, it would clearly help adding custom ignoring logic.
I faced this issue, so I suppose I'm not the only one.

This PR defines a shouldIgnore method at the base Watcher in `Laravel\Telescope\Watchers\Watcher`, and call this method in every watchers. Power is given back to developers to override the method.


Pythagus:
This is similar #310 
This is a weird solution... But a solution.

Closing this PR.

Issue #1392, opened 2023-10-16
Allow ProcessPendingUpdates to be queued on something different than 'default'


NeurohrByteS:
Currently ProcessPendingUpdates is always queued to the 'default' queue. 
https://github.com/laravel/telescope/blob/3a250a44faa89ba4790ad6207060e90f79f49d1e/src/Telescope.php#L673

It would be great if if this could be configured similar to Scout.
https://laravel.com/docs/10.x/scout#queueing

NeurohrByteS:
Thank you for this great library!

driesvints:
If you could send in a PR we could have a look maybe 👍 

> Thank you for this great library!

Thanks!

NeurohrByteS:
@driesvints As requested: https://github.com/laravel/telescope/pull/1393

PR #1393, opened 2023-10-16
feat: allow ProcessPendingUpdates to be queued on something different than 'default'

NeurohrByteS:
# Why has it changed?
+ #1392
+ Currently ProcessPendingUpdates only runs on the "default" queue. However in projects with multiple task workers this is not always desired.

# What has changed?
+ 1x Configuration stub
   + The config key is structured similarly to Laravel Scout's
+ 2x ProcessPendingUpdates dispatches

# How was it tested?
+ Changes are currently active on a two staging servers of the author one with default configuration and one changed to a custom queue worker.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -64,7 +64,6 @@ return [
         'queue' => env('TELESCOPE_QUEUE_QUEUE', 'default'),
     ],
 
-
     /*
     |--------------------------------------------------------------------------
     | Telescope Master Switch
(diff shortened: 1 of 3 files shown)


taylorotwell:

NeurohrByteS:
@driesvints No worries. It's only a minor annoyance for me, which I can easily patch in my projects.
In the big picture I'm glad about the "mother hen" approach towards Laravel, since I rely on its stability too :-)

PR #1396, opened 2023-10-20
Refactor(QueryWatcher): Consider only identical queries as duplicates

lorenzofk:
**Summary**

This pull request addresses a potential issue of defining duplicated queries in the Telescope package. Currently, the definition of a "duplicated" query includes queries with the same structure but different bindings. This PR proposes a change so that only queries that are exactly the same will be considered duplicated. This aligns with a stricter definition of duplicated queries, focusing on identical SQL statements.

**Changes Made**

- Updated the logic in the QueryWatcher to generate the same hash only if they are exactly the same.
- Added new unit tests to ensure that only queries with identical structures and bindings are treated as duplicated.

**Motivation**

The current approach includes queries with similar structures but different bindings as duplicates, which might not accurately reflect redundant database calls. By considering only identical queries as duplicated, we provide a more accurate representation of query duplication, enabling developers to optimize database interactions more effectively.

### DIFF
--- a/src/Watchers/QueryWatcher.php
+++ b/src/Watchers/QueryWatcher.php
@@ -68,7 +68,7 @@ class QueryWatcher extends Watcher
      */
     public function familyHash($event)
     {
-        return md5($event->sql);
+        return md5($this->replaceBindings($event));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
Going to hold off on this for now.

PR #1409, opened 2023-11-15
Add option to suppress prune output

Synchro:
`telescope:prune` is often run from cron, and a long-standing cron convention is that scheduled commands should only produce output when something goes wrong. This helps to reduce noise and alert fatigue. Telescope's prune command produces output stating the number of entries deleted, even if there are none, and no error has occurred. This results in (typically) daily emails being sent to admins that don't actually say anything important. Many commands have an option to suppress non-error output using a `--quiet` or `-q` command line switch, and this PR adds that to telescope's prune command like so:

    php artisan telescope:prune --quiet
    php artisan telescope:prune -q

Side note: when writing the tests, I noticed that there doesn't seem to be an assertion/expect option for there to be no output, so I had to work around that by using `doesntExpectOutput()`

### DIFF
--- a/src/Console/PruneCommand.php
+++ b/src/Console/PruneCommand.php
@@ -12,7 +12,7 @@ class PruneCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'telescope:prune {--hours=24 : The number of hours to retain Telescope data}';
+    protected $signature = 'telescope:prune {--hours=24 : The number of hours to retain Telescope data} {--q|quiet : Suppress output}';
 
     /**
      * The console command description.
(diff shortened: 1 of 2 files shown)


taylorotwell:
This is built into every command already?

Synchro:
Oh, thanks! This was news to me, and it's not in the docs, but I tracked it down. In case anyone else is looking for this, the output verbosity options are defined in the [InteractsWithIO](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Console/Concerns/InteractsWithIO.php#L53) interface, which maps through to the underlying Symfony [OutputInterface](https://github.com/symfony/symfony/blob/6.4/src/Symfony/Component/Console/Output/OutputInterface.php#L21).

PR #1418, opened 2023-12-11
Ignore pulse and livewire

tomswinkels:
Ignore pulse and livewire by default

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -97,6 +97,8 @@ return [
 
     'ignore_paths' => [
         'nova-api*',
+        'livewire*',
+        'pulse*',
     ],
 
     'ignore_commands' => [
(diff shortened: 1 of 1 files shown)


francoism90:
@tomswinkels I don't know if ignoring livewire is a good thing, as this also helps debugging what's get pushed on state changes.

tomswinkels:
@francoism90 you can always enable it. But in default it spams the telescope logging.

AidasK:
Must disagree with this change. Livewire contains mostly developer code and ignoring all of the exceptions there makes no sense. Had an exception for a while in one of livewire methods and haven't noticed it because of this.

Imo it's would be better to ignore all 200 requests instead as this option is too powerful as it ignores exceptions and logs in these routes too.

PR #1454, opened 2024-03-28
Update FormatModel.php

truestealth:
Object of class Illuminate\Database\Query\Expression could not be converted to string


driesvints:
Please add a thorough explanation what this fixes along with an example that triggers the issue. Thanks

truestealth:
@driesvints This is due to https://laravel.com/docs/10.x/upgrade#database-expressions when the telescope cannot process the insertion into the database.

taylorotwell:
No recreation instructions.

PR #1456, opened 2024-04-05
Add ignore paths option for request watcher

niteen1593:
This pull request introduces a new configuration option, `ignore_paths`, to the `RequestWatcher` class in Telescope.

**Motivation:**
Currently, Telescope provides a global configuration option to specify paths to be ignored, but there isn't a way to ignore specific request URIs at the level of the `RequestWatcher` class. Consequently, when we configure `ignore_paths` globally, it applies to all watchers except `QueryWatcher`. This limitation prevents monitoring of other watchers such as `JobWatcher`.

By default, Telescope ignores certain paths like
```php
 'ignore_paths' => [
        'livewire*',
        'nova-api*',
        'pulse*',
    ],
```    
However, this global setting ignores all livewire request so we can't monitor `Event`, `Job` or `Job Batch` dispatched in livewire request.
This pull request resolves this issue by allowing for more granular control over ignored paths within the `RequestWatcher`.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -181,6 +181,7 @@ return [
             'size_limit' => env('TELESCOPE_RESPONSE_SIZE_LIMIT', 64),
             'ignore_http_methods' => [],
             'ignore_status_codes' => [],
+            'ignore_paths' => [],
         ],
 
         Watchers\ScheduleWatcher::class => env('TELESCOPE_SCHEDULE_WATCHER', true),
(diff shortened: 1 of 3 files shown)


taylorotwell:

niteen1593:
Thanks @taylorotwell for reviewing my PR. I have gone through changes made in this PR multiple times And think this will be great option if it will part of framework that will help users to ignore request path directly from framework config instead of pulling one more package to achieve this simple functionality.
Thanks

PR #1466, opened 2024-04-17
add queue connection customization

ariaieboy:
The `ProcessPendingUpdates` job uses the application's default queue and is not customizable.
I've added two new methods `useConnection` and `useQueue` that we can use to customize the queue that we are going to use to dispatch the update job.


taylorotwell:
Can we use a config value instead?

Issue #1468, opened 2024-04-17
Add crossorigin anonymus to the mix configuration because it's not possible to use with a CDN for all the resources


dantes4ur:
### Telescope Version

5.0.1

### Laravel Version

10.48.4

### PHP Version

8.2

### Database Driver & Version

MySQL 8

### Description

Previously, there were no problems of using telescope without this attribute `crossorigin="anonymous"` on `link` or `script` tags. But right now I found a weird problem regarding the script and styles tag if it's consumed from a CDN.

**The error:** 
```
Subresource Integrity:
The resource 'https://cdn.domain.com/vendor/telescope/styles.css' has an integrity attribute, but the resource requires the request to be CORS enabled to check the integrity, and it is not. The resource has been blocked because the
integrity cannot be enforced.
```

**What I found:** 
https://stackoverflow.com/questions/35323268/how-to-solve-resource-requires-the-request-to-be-cors-enabled-resource-has-bee 

I think that is related with the webpack configuration on the production files in the package. 
https://webpack.js.org/configuration/output/

### Steps To Reproduce

Upload your assets to CDN (other domain) and then try to load from a https domain (preferably not local) and then you'll see that your telescope page does not render as expected.

jason-nabooki:
Yeah, we had this issue too but that change has been reverted.

Upgrade to 5.0.3.

dantes4ur:
> Yeah, we had this issue too but that change has been reverted.
> 
> Upgrade to 5.0.3.

I did that, but I found the following problems related to Vite: 

Illuminate\Foundation\ViteManifestNotFoundException: Vite manifest not found at: .../public/vendor/telescope/manifest.json in .../vendor/laravel/framework/src/Illuminate/Foundation/Vite.php:728 

I don't know why it's looking for vite manifest when this is pure webpack. 

https://github.com/laravel/framework/blob/57c6f974654867c2f2c9f82e7e2014669b30b4f1/src/Illuminate/Foundation/Vite.php#L743 this is the line causing error 500, I don't know if there's a workaround because I'm not using assets at all, my project is pure API. 

One thing It would be nice IMO is that class wouldn't look exactly for public_path, sometimes you deploy all the public assets into a S3 or bucket like distribution and use a full ASSET_URL outside your project, wdyt?

jason-nabooki:
Not sure. Didn't have any issues upgrading to `5.0.3`. Only thing I had to do was invalidate the assets on Cloudfront.

[...]

PR #1470, opened 2024-04-18
Add crossorigin anonymus to the mix configuration because it's not possible to use with a CDN for all the resources #1468

dantes4ur:
Fixing #1468 and this is a small fix to enable the telescope to serve from CDN, very useful for distributing public folder into S3/Cloudfront

### DIFF
--- a/resources/views/layout.blade.php
+++ b/resources/views/layout.blade.php
@@ -14,7 +14,7 @@
     <!-- Style sheets-->
     <link rel="preconnect" href="https://fonts.bunny.net">
     <link href="https://fonts.bunny.net/css?family=figtree:300,400,500,600" rel="stylesheet" />
-    <link href="{{ asset(mix($cssFile, 'vendor/telescope')) }}" rel="stylesheet" type="text/css">
+    <link href="{{ asset(mix($cssFile, 'vendor/telescope')) }}" rel="stylesheet" type="text/css" crossorigin="anonymous">
 </head>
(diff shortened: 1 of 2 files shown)


taylorotwell:
A comment on that issue says you just need to update?

dantes4ur:
> A comment on that issue says you just need to update?

Hey Taylor! Big fan of your work right here. 

Found another error around Vite, discussing on the issue I opened

jason-nabooki:
I don't think this is required since `5.0.2` was reverted? We're serving our telescope assets (5.0.3) from a CDN (Cloudfront) just fine without the attribute.

We did, however, run into issues with `5.0.2` when there was an `integrity` attribute added (https://github.com/laravel/telescope/pull/1461). The `crossorigin="anonymous"` would have helped there but for `5.0.3` there shouldn't be any issues?

dantes4ur:
> I don't think this is required since `5.0.2` was reverted? We're serving our telescope assets (5.0.3) from a CDN (Cloudfront) just fine without the attribute.
> 
> We did, however, run into issues with `5.0.2` when there was an `integrity` attribute added (#1461). The `crossorigin="anonymous"` would have helped there but for `5.0.3` there shouldn't be any issues?

As I pointed out here https://github.com/laravel/telescope/issues/1468#issuecomment-2062914816 I found issues with Vite on version 5.0.3, what would be a workaround?

PR #1472, opened 2024-04-21
CS Fixing to align with the default laravel preset for Pint

alidevweb:
Using Laravel Pint with the default `laravel` preset, then installed Telescope, but Pint suggest these changes.
I think the Laravel packages should follow the Laravel coding standards by default. So Pint will not be angry about generated files by Laravel packages.


driesvints:
Thanks!

PR #1484, opened 2024-06-07
Remove column name for uuid

lazypolymath:
The default column name for the uuid column type is uuid.


lazypolymath:
Didn't see 5.x supports Laravel <= 8. My bad.

Issue #1435, opened 2024-02-05
query with an escaped question mark does not reflect correct bindings


WaqasIbrahim:
### Telescope Version

4.17.5

### Laravel Version

10.43.0

### PHP Version

8.1.21

### Database Driver & Version

PostgreSQL 15.3 (Debian 15.3-1.pgdg120+1)

### Description

Hello, I have a Query builder macro in my application that checks if a JSON column contains any of the provided values. There is no built-in laravel function for this so I use a raw statement for this.

```php
Builder::macro('whereJsonContainsAny', function (string $column, array $values) {
    /** @var Builder $this */
    $placeholders = implode(',', array_map(fn () => '?', $values));

    return $this->whereRaw("{$column} ??| array[{$placeholders}]", $values);
});
```

`?|` operators is used checks if any of the specified values exist in jsonb array.
`??|` in this is used to escape `?` so query builder does not view this as a binding placeholder. 

The query is executed successfully but telescope does not report the correct query.

Telescope result:

```sql
select
  "id"
from
  "movies"
where
  "release_date" between '2023-02-05 00:00:00'
  and '2024-02-05 00:00:00'
  and ("categories") :: jsonb @ > '["action"]'
  and categories 'thriller''crime' | array ['adventure','one-man-army','one-person-army-action']
  and keywords 'spy''terrorist' | array ['mission','secret-agent','action-hero','assasin','intelligence-service',?,?,?,?]
order by
  "rank" desc
limit
  21 offset 0
```

Expected result:

```sql
select
  "id"
from
  "movies"
where
  "release_date" between '2023-02-05 00:00:00'
  and '2024-02-05 00:00:00'
  and ("categories") :: jsonb @ > '["action"]'
  and categories ?| array ['thriller', 'crime', 'adventure']
  and keywords ?| array ['one-man-army','one-person-army-action','spy','terrorist','mission','secret-agent','action-hero','assasin','intelligence-service']
order by
  "rank" desc
limit
  21 offset 0
```

Another minor issue is the space between `@>` which throws an error if you copy and execute the query directly.

### Steps To Reproduce

Wrtie a raw statement using an `?|` operator comparing a jsonb array column for postgresql.

driesvints:
A fix will be released today. Thanks!

PR #1486, opened 2024-06-14
[5.x] Enhance query preview when using the pgsql driver

morloderex:
This is an attempt to fix the rendering issue that was reported in #1435

It does it by simply using the laravel framework build-in sql formatting logic if the laravel version is above 10.x. And appending the driver name to telescope content so that the `sql-formatting` javascript formatting library can format it in the correct dialect, while still falling back to use default of sql if it is not present.

I have also updated the javascript highlighter to it's own component in order to detect the correct dialect for psgql to take place.

However there is a breaking change with doing it in this way due to the framework's build-in sql formatting logic doesn't support named bindings. But given that framework's query builder itself only supports `?` parameters anyway i don't really see that as a huge issue.


driesvints:
Thanks for this PR @morloderex!

Issue #786, opened 2019-12-10
Display which queries are duplicated


MatanYadaev:
It's great that now Telescope is displaying count of the duplicated queries.

**The next step, in my opinion, is to display which of the queries are duplicated.**

I have lots of queries that being done in each request/job, and they all are very long, it's a bit hard to say which of these are duplicate.
It would be nice to have some indicatior on each query that representing a "duplicated query", or maybe a filter button that on click displays only the duplicated queries.

driesvints:
Same as https://github.com/laravel/telescope/issues/603 I think.

themsaid:
Indeed a duplicate. Feel free to open a PR.

Issue #603, opened 2019-03-29
Duplicate Query Indication


JonPurvis:
Apologies in advance if it already exists and I've just missed it. I was wondering if it's possible to have an indication of duplicate queries in the list of queries when viewing a request. It'd just make it easier to see at a glance which ones are duplicates.

danilopolani:
I don't think currently there's a way, also because of the way data is saved in `telescope_entries` it could be a bit tricky. Because MySQL introduced `REGEXP_SUBSTR` only since 8.0, a valid but not precise (it doesn't include query bindings) query to achieve it also from previous versions could be:

### TL;DR:

Use the query after the line and there you go. I think it would take a bit of time to implement this on Telescope because of displaying single query detail / expand the aggregated data.

### LR:

```
SELECT
	COUNT(sequence) AS cnt,
	SUBSTRING_INDEX(SUBSTRING_INDEX(content, '"sql":"', -1), '","time":', 1) AS q
FROM telescope_entries
	WHERE type = 'query'
GROUP BY q
ORDER BY cnt DESC
```

Just to explain the query a bit:

- In the `telescope_entries` table, queries are stored in a JSON with the key `sql`. With `SUBSTRING_INDEX(content, '"sql":"', -1)` we get the content from the first apis opening the query (like `"sql":"select * from ...",`) to the end of the JSON. With `-1` you get the "right content" of the matching part.
- With the second substring - `SUBSTRING_INDEX(.., '","time":', 1)`, we get the content from `","time":`  , namely the key after `sql`, to the _start_ of the JSON (in this case, the query itself). With `1` instead of `-1` we get the "left content" of the match.
- Finally, with `WHERE type = 'query'` we get only query logs (because there may be also requests, cache, redis etc.) 

The rest is just common MySQL stuff like `GROUP BY`, `COUNT` and so on. I know I didn't explain it very well, but trust me haha. 

----

We can also recycle the logic above used to extract the query, to extract the bindings and, merging everything together, we can get a query to have real unique queries counter:

```
SELECT
	COUNT(sequence) AS cnt,
	CONCAT(
		SUBSTRING_INDEX(SUBSTRING_INDEX(content, '"bindings":', -1), ',"sql":', 1), # extract bindings
		'::', # just a fancy separator, maybe to change for performance?
		SUBSTRING_INDEX(SUBSTRING_INDEX(content, '"sql":"', -1), '","time":', 1) # extract query
	) AS full_query
FROM telescope_entries
[...]

PR #1496, opened 2024-07-16
Add duplicated query indicator

nicDamours:
## Summary
Added a badge to indicate which queries were duplicated, in the "Related Queries" tabs. The badge can be clicked on, to scroll to the first instance of the query, which will be highlighted. The first instance of a duplicated query is not marked with a badge, since it's not technically duplicated yet. 

## Screenshots
Light theme:
image

Dark theme:
image

An example of the badge with a badge for a long query:
image

## Motivation
This feature came from a need to know which queries were duplicated, when looking at the queries for a given http request. Knowing exactly which queries are duplicated can be very useful when attempting to optimize a request with eager loading. Also, know which instance of the query is the first one can help to identify when this query come from and how to prevent further duplication. 

There was quite a few issue regarding this problem, and even a few attempt, but it never got fixed. So here I am, throwing my solution into the ring. 

Here are a list of the related issue 
#603 
#786 
#1326 

And a PR to fix the issue
#1328 

## Testing
I wanted to write a complete test suite for this feature. However, since this is a front-end only fix, I could not use the regular phpunit framework. I would have needed to install jest, and setup everything needed to even be able to run a single test. I felt like this was not my place, and I did not want to do more than necessary on that feature.

If, at some point, a testing framework is added to the front-end part of telescope, I will gladly come back and write some tests.

### DIFF
--- a/resources/js/components/RelatedEntries.vue
+++ b/resources/js/components/RelatedEntries.vue
@@ -27,8 +27,15 @@
          */
         mounted(){
             this.activateFirstTab();
+            this.registerDocumentClickListener();
         },
 
+        /**
+         * Remove the previously added listeners
+         */
+        beforeUnmount() {
+            this.unregisterDocumentClickListener()
+        },
 
         watch: {
             entry(){
@@ -94,6 +101,15 @@
                 if(entryElement) {
                     entryElement.scrollIntoView({ behavior: "smooth", block: "center", inline: "center"});
                 }
+            },
+            blurHighlightedEntry() {
+                this.highlightedEntry = null;
+            },
+            registerDocumentClickListener() {
+                document.addEventListener('click', this.blurHighlightedEntry)
+            },
+            unregisterDocumentClickListener() {
+                document.removeEventListener('click', this.blurHighlightedEntry)
             }
         },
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Going to hold off on this one for now.

nicDamours:
@taylorotwell Would you mind sharing why? I feel like this would make a great addition to Telescope. As a matter of facts, I've been running this exact patch for about a year, on my personal installation of telescope and it made debugging a lot easier.

emaadali:
+1 for including this feature

mattpannella:
chiming in to ask why this keeps being shut down. been looking for this feature for years and every time i look it up, i end up seeing closed issues or PRs like this one

waterloomatt:
> @taylorotwell Would you mind sharing why? I feel like this would make a great addition to Telescope. As a matter of facts, I've been running this exact patch for about a year, on my personal installation of telescope and it made debugging a lot easier.

I'd like to do that too. What steps do you have to do after copy/pasting the changes from this PR into the vendor/laravel/telescope/resources/js/components/RelatedEntries.vue and vendor/laravel/telescope/resources/sass/base.scss files?

nicDamours:
@waterloomatt From memory, you have to install the dependencies at the root of the `vendor/laravel/telescope`:

```
npm install
```

Then, you have to compile the assets using the command, at the root of the directory `vendor/laravel/telescope` : 

```
npm run production
```

Finally, you have to republish the compiled assets using the command, at the root of your project.

```
php artisan telescope:publish
``` 

You might have to clear your cache and do a hard refresh, before seeing any changes.

I'm recalling this from memory, so it might not be accurate. I'll test it out tomorow and update my comment if need be.

PR #1499, opened 2024-07-31
add support for non-standard PDOs

macbookandrew:
I’m using a mysql database for my app, as well as this package to access a FileMaker database for some additional data: https://github.com/gearbox-solutions/eloquent-filemaker

That package is firing a `\Illuminate\Database\Events\QueryExecuted` event for logging purposes.

However, that results in `ERROR  Call to a member function quote() on string in vendor/laravel/telescope/src/Watchers/QueryWatcher.php on line 122.` in Telescope because [this line](https://github.com/laravel/telescope/blob/a678a14018f27df566b5478804b530fdf618c812/src/Watchers/QueryWatcher.php#L122) is trying to get a `PDO` instance. Because FileMaker databases are accessible via a REST API and not an actual database connection, there is no PDO instance available.

This PR addresses that by detecting whether the app has an actual `PDO` instance before trying to call the `quote()` method.

### DIFF
--- a/src/Watchers/QueryWatcher.php
+++ b/src/Watchers/QueryWatcher.php
@@ -119,7 +119,11 @@ class QueryWatcher extends Watcher
     protected function quoteStringBinding($event, $binding)
     {
         try {
-            return $event->connection->getPdo()->quote($binding);
+            $pdo = $event->connection->getPdo();
+
+            if ($pdo instanceof \PDO) {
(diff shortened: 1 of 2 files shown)


driesvints:
This seems to break the test suite

macbookandrew:
@driesvints I just pushed a fix for the failing tests…would you mind re-considering?

Issue #1410, opened 2023-11-18
Prune command fails to delete records then using with mysql on planetscale


Arturas0:
### Telescope Version

4.17.2

### Laravel Version

10.31.0

### PHP Version

8.2

### Database Driver & Version

MySQL on planet scale

### Description

After changing db from MySQL to MySQL on planet scale command `php artisan telescope:prune` fails, then table `telescope_entries` has more than 100k records. Planet scale has limits (https://planetscale.com/docs/reference/planetscale-system-limits) and one of them is `Per-query rows returned, updated, or deleted` 100k.

### Steps To Reproduce

1. Use planet scale as DB;
2. Insert 101k records to table `telescope_entries`;
3. Run command `php artisan telescope:prune`

driesvints:
But deletes for prunes are already chunked by 1000 items each so we never delete 100K records at once? So I don't understand how this is possible?

https://github.com/laravel/telescope/blob/4.x/src/Storage/DatabaseEntriesRepository.php#L352

Arturas0:
Chuncks are used to delete records, but before deletion, data is selected.
If I am correct, than this query ` $query = $this->table('telescope_entries')->where('created_at', '<', $before)` translates to SQL which tries to select data and causes an error.

![Screenshot_1](https://github.com/laravel/telescope/assets/29736880/282fce8d-341c-42b3-92b4-f3d382bf3900)

driesvints:
No, that's just a query builder instance. `->take()` is applied to it afterwards which limits the result to 1000.

Arturas0:
Thank you for clarification. I tried to recreate issue, I installed telescope for new project with db planet scale, after inserting more than 100k records into table `telescope_entries` command `php artisan telescope:prune` worked correctly, so it looks like I was incorrect prune command works as expected.

But I had problem with telescope button `clear entries` than table has more than 100k records.
<img width="275" alt="Screenshot 2023-11-19 at 14 58 47" src="https://github.com/laravel/telescope/assets/29736880/eafe4f5c-0649-4e51-9941-d0c9282acadd">

After clicking this button I can see query was executed:
<img width="1360" alt="Screenshot 2023-11-19 at 14 55 44" src="https://github.com/laravel/telescope/assets/29736880/702a2ce8-5cc7-4bc8-8fc5-dd7004c8622c">

And then I saw error from planet scale, after manually deleting some records, to have less than 100k query was executed successfully:
[...]

PR #1506, opened 2024-08-19
[5.x] Enhance Telescope Clear Method

a1383n:
This PR refines the `clear` method in the Laravel Telescope `DatabaseEntriesRepository` for a more efficient and reliable table-clearing process.

#### Key Improvements:
- **Safe Truncate:** The method now attempts to truncate the `telescope_entries` and `telescope_monitoring` tables using `Schema::withoutForeignKeyConstraints`, effectively managing foreign key constraints.
- **Fallback to Chunked Deletion:** If truncation fails due to a `PDOException`, the method automatically falls back to chunked deletion, ensuring that the clearing process completes smoothly.
- **Performance Boost:** Truncation is significantly faster than chunked deletion, optimizing the clearing process while preserving fallback options for reliability.

#### References:
- This PR resolves issues introduced in #991, where truncation without handling foreign keys caused failures.
- It maintains the chunked deletion logic added in #1415 and #1410 to support databases with row deletion limits.

This solution is stable, backward-compatible, and ready for review.

### DIFF
--- a/src/Storage/DatabaseEntriesRepository.php
+++ b/src/Storage/DatabaseEntriesRepository.php
@@ -6,6 +6,7 @@ use DateTimeInterface;
 use Illuminate\Database\UniqueConstraintViolationException;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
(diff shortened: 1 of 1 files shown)


driesvints:
Tests are failing here. Please resend this with passing tests, thanks.

Issue #1476, opened 2024-05-07
Preventing migrations from running in Production


Rydgel:
I'm upgrading from v4 to v5, and it is said that:

> Telescope 5.0 no longer automatically loads migrations from its own migrations directory. Instead, you should run the following command to publish Telescope's migrations to your application:

I'm currently using the local only install like this:

```php
public function register(): void
{
    if ($this->app->environment('local')) {
        $this->app->register(\Laravel\Telescope\TelescopeServiceProvider::class);
        $this->app->register(TelescopeServiceProvider::class);
    }
}
```

If I understand correctly, I am now forced to publish the migrations.
That would make them run on production after a deploy, is there a way around it?

driesvints:
Heya, not at this time I'm afraid. But the above will still prevent Telescope from running in your app on production. Only the migrations will be run.

Rydgel:
thanks for the quick response

BobbyBorisov:
@Rydgel hey what did you end up doing?

PR #1508, opened 2024-08-22
Skip migrations when Telescope is disabled

kauhat:
Hi,

My project uses an in-memory SQLite database for testing in CI, and `RefreshDatabase` tests stopped working after upgrading to v5 and publishing the Telescope migration file. In my case, the tests were failing as the `telescope.storage.database.connection` config value had not been updated to point to the SQLite database.

This change skips Telescope migrations when the `TELESCOPE_ENABLED` flag is false, such as when running tests using the standard Laravel PHPUnit config. However, it's not a great solution as migrations will sometimes need to be re-run after enabling Telescope.

Related:
https://github.com/laravel/telescope/issues/1476

### DIFF
--- a/database/migrations/2018_08_08_100000_create_telescope_entries_table.php
+++ b/database/migrations/2018_08_08_100000_create_telescope_entries_table.php
@@ -19,6 +19,10 @@ return new class extends Migration
      */
     public function up(): void
     {
+        if (! config('telescope.enabled')) {
+            return;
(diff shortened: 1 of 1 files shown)


driesvints:
These migrations are published to your app. You can modify them how you like 👍

kauhat:
That's fair enough. Thanks for your quick response!

PR #1516, opened 2024-09-08
✨ feat: sample rate added

mhkarimi1383:
Added a sample rate config and `isEnabled` function to handle sample rates this will make this package more usable in bigger scales

Also for safety and to make it more reusable I made that set that telescope if it's enable or not in the context of request

If needed I can Also make `isEnabled` function customizable

> NOTE: I'm not a PHP/Laravel dev so some mistakes in changes are possible

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -20,6 +20,18 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Telescope Sample Rate
+    |--------------------------------------------------------------------------
+    |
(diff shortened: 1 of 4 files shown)


mhkarimi1383:
Context is only available is laravel 11, what we have to do for older version?

taylorotwell:

mhkarimi1383:
@taylorotwell 
But this feature is a needed thing for a tool like this one, since other options have this feature and they are much more scalable

PR #1519, opened 2024-09-11
[5.x] Added 2 types search for entries

merkulovka:
This PR adds two new entries searches:
1. By email;
2. By response status. 

While testing our products, we were faced with the fact that records are added too quickly and it becomes very difficult to catch the desired record, the solution was to add an email search

telescope


taylorotwell:

PR #1537, opened 2024-10-17
can change table names in config

VirusEcks:
can use custom names for the 3 telescope tables

in config they have the default values
but in .env can use these 3 names

```
TELESCOPE_TABLE
TELESCOPE_TAGS_TABLE
TELESCOPE_MONITORING_TABLE
```


taylorotwell:

greal:
@taylorotwell This would be a very useful improvement for me, I need to set up prefixes for tables, the absence of this function is a bit strange, there is no flexibility in customizing this tool.

Issue #1479, opened 2024-05-10
Telescope Jobs watcher causes queued job  exception


Yahav:
### Telescope Version

Latest

### Laravel Version

10

### PHP Version

8.2

### Database Driver & Version

MariaDB

### Description

When the Job watcher is enabled, jobs that are deleting the model they receive are never marked as Done nor the JobProcessed event dispached. 
The problem is that Telescope listens to the JobProcessed event and try to fetch the model, however the model was already deleted and so a ModelNotFoundException is thrown causing the job to fail without being marked as failed or done.
Important to say that the job was actually completed but was never marked as such because of this issue.

Illuminate\Database\Eloquent\ModelNotFoundException No query results for model [App\Models\User]. 
    vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:621 Illuminate\Database\Eloquent\Builder::firstOrFail
    vendor/laravel/framework/src/Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php:109 App\Jobs\DeleteUser::restoreModel
    vendor/laravel/framework/src/Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php:62 App\Jobs\DeleteUser::getRestoredPropertyValue
    vendor/laravel/framework/src/Illuminate/Queue/SerializesModels.php:93 App\Jobs\DeleteUser::__unserialize
    [internal] unserialize
    vendor/laravel/telescope/src/Watchers/JobWatcher.php:247 Laravel\Telescope\Watchers\JobWatcher::getCommand
    vendor/laravel/telescope/src/Watchers/JobWatcher.php:213 Laravel\Telescope\Watchers\JobWatcher::updateBatch
    vendor/laravel/telescope/src/Watchers/JobWatcher.php:106 Laravel\Telescope\Watchers\JobWatcher::recordProcessedJob
    vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:458 Illuminate\Events\Dispatcher::Illuminate\Events\{closure}
    vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:286 Illuminate\Events\Dispatcher::invokeListeners
    vendor/laravel/framework/src/Illuminate/Events/Dispatcher.php:266 Illuminate\Events\Dispatcher::dispatch
    vendor/fntneves/laravel-transactional-events/src/Neves/Events/TransactionalDispatcher.php:110 Neves\Events\TransactionalDispatcher::dispatch
[...]

PR #1539, opened 2024-10-23
[5.x] Fix `JobWatcher` when processing deleted instance of `SerializesModels`

hapidjus:
If there is a failure when unserializing the command, (for instance if the job deleted a model it received,  of the database connection is set dynamically) the jobs will never proceed to processed. See #1479

In this PR unserialize is removed in favor of preg_replace to extract the batch_id


crynobone:
@hapidjus shouldn't we use the new approach only if `unserialize()` return an exception?

hapidjus:
> @hapidjus shouldn't we use the new approach only if `unserialize()` return an exception?

Makes sense. 
This PR was mainly to confirm that using preg_match is a reasonable approach here.
The PR currently doesn’t handle encrypted data or include any tests.

However I'm not sure we can catch exceptions that happens "inside" the unserialize function. It's not necessarily unserialize() that throws.

hapidjus:
I've updated the PR now to try to unserialize first and only use preg_match as fallback.

crynobone:
@hapidjus Thanks, I'll review it and possibility add some tests to cover the new use case. Will mark it as ready when I'm done with it

src/Watchers/JobWatcher.php
```diff
@@ -266,7 +267,7 @@ protected function getBatchId(array $data)
             $properties = ExtractProperties::from($command);
 
             return $properties['batchId'] ?? null;
-        } catch (\Exception|\Error) {
```
hapidjus: I understrand that `catch (\Exception|\Error)` is a bit greedy. But there are other ways that the unserialize can go wrong. For instance when in a multi tenancy context and the database connection has not been set yet. See this issue for instance https://github.com/archtechx/tenancy/issues/1263

crynobone: If there a use case to include other exceptions we need to see the actual use case and replication steps so we can add tests to cover that usage. Having the scope too wide add a huge risk to this PR getting rejected.

hapidjus: Alright, I'll see if I can make a minimal reproduction of it. Bur for now I'm happy to have this merged and make another PR later.

PR #1550, opened 2024-11-23
[5.x] fix implicitly marking parameter as nullable

SupianIDz:
`Laravel\Telescope\Storage\DatabaseEntriesRepository::__construct()`: Implicitly marking parameter $chunkSize as nullable is deprecated, the explicit nullable type must be used instead


SiddharthPant:
First thanks @SupianIDz for already raising PR for an issue I am also facing in one of my projects after moving to PHP v8.4. Second, could someone help me understand why the error for this appears in Pail's tailed logs but not in `laravel.log` file? Is there a way even errors like these can be captured in `laravel.log` file?

Issue #542, opened 2019-01-24
Feature: Hide menu item when watcher is turned off


tomswinkels:
Hide the menu item when we turned a watcher off.

themsaid:
No plans for this at the moment :)

ScriptingBeating:
Or at least give an option to publish telescope views. So that we can modify it.

PR #1552, opened 2024-12-01
Add option to hide menu items for disabled watchers

rickiewars:
Implementation for [this feature request](https://github.com/laravel/telescope/issues/542)

I have some watchers disabled because I do not use them that often.
I would like to be able to hide the respective navigation items.
This pull request implements that.
It makes the sidebar a bit less cluttered if you only use a handfull of watchers.

I've implemented this feature as an opt-in feature using a config parameter.
Only if the `hide_disabled_watcher_pages` config is true and if the corosponding watcher is disabled, the menu item will be hidden.

Please let me know if anything needs to be changed.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -121,6 +121,8 @@ return [
         //
     ],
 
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #1578, opened 2025-03-25
[5.x] Record plain/text request content

### LINKED ISSUE #1569: Request payload of Content-Type: text/plain not shown
### Telescope Version

5.5.0

### Laravel Version

11.34.2

### PHP Version

8.3

### Database Driver & Version

_No response_

### Description

When sending an http post request with Content-Type: text/plain, the RequestEvent Payload is `[]` instead of showing the request content.

### Steps To Reproduce

`curl -X POST http://localhost/api -H "Content-Type: text/plain" -d "Some plain text test"`

results in a request event with an empty payload:
Image

cosmastech:
To close https://github.com/laravel/telescope/issues/1569

Requests were predicated on being a JSON body or form input. Now, if we receive a payload with content-type: text/plain, it will record the body.

Before:
image

After:
image

### DIFF
--- a/src/Watchers/ClientRequestWatcher.php
+++ b/src/Watchers/ClientRequestWatcher.php
@@ -199,8 +199,8 @@ class ClientRequestWatcher extends Watcher
             return $request->data();
         }
 
-        if ($this->hasHeader('Content-Type') &&
-            str_contains($this->header('Content-Type')[0], 'text/plain')) {
+        if ($request->hasHeader('Content-Type') &&
+            str_contains($request->header('Content-Type')[0], 'text/plain')) {
 
             return $request->body();
         }
--- a/src/Watchers/ClientRequestWatcher.php
+++ b/src/Watchers/ClientRequestWatcher.php
@@ -155,11 +155,15 @@ class ClientRequestWatcher extends Watcher
     /**
      * Format the given payload.
      *
-     * @param  array  $payload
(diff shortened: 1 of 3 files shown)


kohlerdominik:
@cosmastech Thanks very much for the fix, greatly appreciated for our use-case.

Have you thought instead of just hard require `text/plain` header for your fix, inverting the condition (checking if it's a handled array-like as `application/json` or `multipart/form-data`), and fallback and assume it's at least text-compatible if it's not an explicit known arrayable? 

This would add support for more text formats like `text/xml`, `text/csv`, `text/yaml`, `application/xml`, `image/svg+xml` and many more implicit.

cosmastech:
> @cosmastech Thanks very much for the fix, greatly appreciated for our use-case.

You're very welcome. I had the exact same problem testing SNS webhooks 😅 

> Have you thought instead of just hard require `text/plain` header for your fix, inverting the condition (checking if it's a handled array-like as `application/json` or `multipart/form-data`), and fallback and assume it's at least text-compatible if it's not an explicit known arrayable?
> 
> This would add support for more text formats like `text/xml`, `text/csv`, `text/yaml`, `application/xml`, `image/svg+xml` and many more implicit.

That's definitely a good point. I probably won't have time to dedicate to PRing that change, but I would say that I could see value in it.

PR #1583, opened 2025-04-09
Added isJob() for consistent entry type checks

DavydeVries:
## Add `isJob()` method usage for entry type checks

Replaced the direct type comparison:

```php
if ($entry->type === EntryType::JOB)
```

with a more expressive and consistent method call:

```
if ($entry->isJob())
```

This aligns with existing patterns like `isRequest()` and improves code readability and consistency.

### DIFF
--- a/src/IncomingEntry.php
+++ b/src/IncomingEntry.php
@@ -253,6 +253,16 @@ class IncomingEntry
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's fine for now.

PR #1474, opened 2024-04-23
[5.x] Readd vite support

mmachatschek:
This applies the same hotfile workaround https://github.com/laravel/horizon/pull/1421 for users that start their laravel application with vite in dev/hot mode

PR #1598, opened 2025-06-04
Migrate to Vite

nckrtl:
This PR migrates from mix to Vite. I created a new PR as https://github.com/laravel/telescope/pull/1474 is open for already a year. Felt easier to start with a fresh slate and it allowed me to get a better feeling of the project structure. 

I tried to mimic Horizon's project structure as much as possible. Copied the same stuff over like `.gitattributes`, `style.ci` etc. Also the way assets are handled are similar to Horizon, which is a [remark](https://github.com/laravel/telescope/pull/1474#issuecomment-2193964673) in the other PR.

If there is a preference to update the other PR instead, I can give that a try as well.


nckrtl:
Bit unsure about styleci. I'd expect it to commit style preferences, but it only reports them. Tried to fix them but its now reporting about an expected trailing comma at `resources/css/styles-dark.scss` while it was not expecting one for `resources/css/styles.scss`

taylorotwell:
Thanks!

PR #1602, opened 2025-06-19
feat: keep Telescope sidebar visible on scroll using sticky positioning

EvanSchleret:
This PR improves the UX of Telescope by making the sidebar navigation sticky during page scroll.

Tested on current 5.x branch.

Let me know if you'd prefer a `fixed` approach instead, but this keeps it fully scoped within the layout flow.


taylorotwell:

EvanSchleret:
Thanks for your reply @taylorotwell. However, this answer leaves me perplexed. What do you mean by "to preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include." ?

My PR only adds three CSS properties to improve, in my opinion, the UX of telescope. Is this a somewhat generic or automatic answer?

Thank you for your possible response and for what you have brought to my life as a web developer.

PR #1607, opened 2025-07-14
Enable logging only slow requests

xwiz:
This enables logging of only requests which take long time to complete but not necessarily from slow queries


taylorotwell:

xwiz:
Hi @taylorotwell  greetings.., not sure what the comment means if you have a different way you'd prefer it's implemented do let me know, it's a feature I would really appreciate for quick debugging when I don't want to use external packages in moderate production environments and I'm sure it will help some others for improving application performance without having to rely on sentry and the likes.

xwiz:
For anyone else needing this, realized this can be easily implemented, added a doc PR

https://github.com/laravel/docs/pull/10623

PR #1616, opened 2025-07-24
[5.x] Fix compile assets

crynobone:


mshamaseen:
hmmm, I'm trying to republish the assest after this change, but getting:
   INFO  No publishable resources for tag [telescope-assets].

PR #1618, opened 2025-07-29
Feature Request: Add "Duplicate Queries" Tab to Highlight Repeated SQL

dharitmn:
### Description

I'd like to propose adding a **"Duplicate Queries"** tab (or section) within each Request preview screen in Laravel Telescope.

This tab would highlight SQL queries that are executed **multiple times during a single request** (i.e., duplicated queries). The purpose is to help developers quickly identify performance bottlenecks such as **N+1 query issues** or **repeated DB hits**.

---

### Why It's Needed

- Telescope currently allows us to inspect all queries with hover previews.
- However, for long SQL queries, the hover view is truncated and hard to compare visually.
- In large requests, **identifying repeated queries manually is inefficient and error-prone**.

A dedicated tab for duplicated queries would provide a much clearer and faster insight into inefficient database usage.

---

### Proposed Functionality

- **Group queries** by the raw SQL string (ignoring bindings or whitespace).
- **Show only those** repeated more than once.
- Add a new tab in the Request detail screen: `Duplicate Queries`
- Display:
  - The query text
  - How many times it was run
  - Optional: Total duration of all executions

---

### Example Scenario

A request runs the same query 6 times:
```sql
SELECT * FROM `users` WHERE `id` = ?
```

Here is a screenshot: 

image

I'm happy to help contribute or collaborate on a pull request if this feature is accepted.


taylorotwell:

Issue #1603, opened 2025-06-26
Interception of "streaming" responses causes issues


avengerweb:
### Telescope Version

5.9.1

### Laravel Version

12

### PHP Version

8.1

### Description

Whenever there is "streaming" response or text/event-stream response telescope intercepts it leading to response being empty for both telescope and whoever made request call.
Such requests are made by for example GenAI libraries (ex. prismphp)

### Steps To Reproduce

You can create dummy api for SSE, some article to implement it in laravel:
https://medium.com/@ahmedm3bead/implementing-server-sent-events-sse-in-laravel-a01da635485b

Then call it using standard laravel Http client, if telescope enabled you will receive no response.

crynobone:
Since most event stream may contents a large amount of content I'm not sure if there's a better alternative here.

crynobone:
Fixed via #1621

PR #1621, opened 2025-08-03
[5.x] Fix Streaming Issue

michaelnabil230:
This PR resolves an issue that occurs when making streamed HTTP requests using `Http::withOptions(['stream' => true])` in Laravel.

🔗 **Fixes:**

* [laravel/telescope#1603](https://github.com/laravel/telescope/issues/1603)
* [prism-php/prism#381](https://github.com/prism-php/prism/issues/381)

### ✅ Changes

* Prevents Telescope from interfering with streamed responses.
* Ensures chunked HTTP responses are handled correctly and without buffering.

This improves the stability and reliability of stream-based operations in Laravel applications.

---

### 🧪 Testing Note

Currently, I haven’t included an automated test for this change, as it's not clear how to effectively test streamed HTTP responses in this case.
The fix has been manually verified and works as expected.

If anyone has experience or ideas on how to write a proper test for streamed requests, contributions are very welcome!

### DIFF
--- a/src/Watchers/ClientRequestWatcher.php
+++ b/src/Watchers/ClientRequestWatcher.php
@@ -97,13 +97,15 @@ class ClientRequestWatcher extends Watcher
      */
     protected function response(Response $response)
     {
-        $content = $response->body();
-
         $stream = $response->toPsrResponse()->getBody();
 
         if ($stream->isSeekable()) {
             $stream->rewind();
+        } else {
+            return 'Stream Response';
         }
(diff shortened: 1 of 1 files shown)


michaelnabil230:
The issue occurs when calling `$response->body()` on a streamed response — it fails to convert to a string, though I'm not entirely sure why. However, adding a condition to check if the response is streamed before accessing the body resolves the problem.

PR #1638, opened 2025-09-19
feat: add N+1 detector watcher

ma32kc:
### Summary
Adds a new NPlusOneWatcher that detects repeated queries sharing the same normalized SQL pattern from the same caller within a short time window and records a warning log entry.

### Motivation
N+1 database patterns are a common performance issue. Telescope already surfaces performance via the Query watcher (e.g., "slow" queries). This watcher complements it by flagging N+1 at the point of occurrence. See the Query watcher docs for option style and prior art.

### Implementation
- New file: src/Watchers/NPlusOneWatcher.php
- Config entry in config/telescope.php with `enabled`, `threshold`, `window`
- Test: tests/Watchers/NPlusOneWatcherTest.php

### BC Breaks
None. The watcher is additive and can be disabled via config.

### Tests
Pass locally with `vendor/bin/phpunit` using the package's Testbench setup.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -206,9 +206,9 @@ return [
         Watchers\ViewWatcher::class => env('TELESCOPE_VIEW_WATCHER', true),
 
         Watchers\NPlusOneWatcher::class => [
-            'enabled'   => env('TELESCOPE_NPLUSONE_WATCHER', true),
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #1648, opened 2025-10-15
Fix/5.x indexing

imhayatunnabi:
Telescope was missing composite indexes for common queries, which leading to:

1. Slow dashboard loading times
2. Poor performance with large datasets
3. Inefficient related entries queries
4. Slow exception grouping

Test Results:

> Dashboard Queries: 1.08ms average (target: <200ms)
> Related Entries: 2.03ms average (target: <100ms)
> Exception Grouping: 1.03ms average (target: <50ms)
> Tag-based Filtering: 0.16ms average (target: <100ms)
> Large Dataset (10,000 entries): 2.29ms (target: <500ms)


taylorotwell:
9 indexes on a table with barely that many columns strikes me as a bit of a smell.

PR #1650, opened 2025-10-21
feat: add shouldIgnoreHosts method to filter HTTP client requests by host

artengin:
Added new functionality to allow configuring host-based ignore rules (e.g., CDNs) in the ClientRequestWatcher.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -143,7 +143,10 @@ return [
             'ignore' => [],
         ],
 
-        Watchers\ClientRequestWatcher::class => env('TELESCOPE_CLIENT_REQUEST_WATCHER', true),
+        Watchers\ClientRequestWatcher::class => [
+            env('TELESCOPE_CLIENT_REQUEST_WATCHER', true),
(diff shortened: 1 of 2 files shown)


artengin:
Hi there,

Our team uses Telescope across all our projects, and sometimes we run into an issue where the HTTP Client watcher tracks repeated CDN requests. There can be quite a few of these requests, which causes some minor inconveniences for us.

I’ve added the ability to filter hosts for HTTP Client tracking, and I hope this can be useful. Our team would really appreciate it if a similar feature could be added directly to Telescope in the future.

Thank you so much!

taylorotwell:
The other watchers have an `enabled` flag when they support other options, but this one doesn't seem to?

artengin:
Yes, it’s actually possible to disable the `ClientRequestWatcher`, so I’ve updated my PR to fix that.
Unfortunately, there’s no way to configure a filter yet, which is why I suggested these changes.

I’ll be happy to apply any further recommendations or adjustments as soon as possible.
Thanks!

Issue #1654, opened 2025-10-31
Crash saving a model with a composite key - Only when telescope is enabled


gbalduzzi:
### Telescope Version

5.5

### Laravel Version

12.28

### PHP Version

8.2

### Database Driver & Version

MySQL 8

### Description

As already reported in #1453, saving a model with a composite key results in a telescope error, when enabled.

The issue was closed with the reasoning that eloquent does not support composite keys, **but the same code works when telescope is not enabled**. 

See the code samples in the steps to reproduce, they work perfectly until telescope is enabled.

### Steps To Reproduce

```
class Test extends Model {
  protected $primaryKey = ['first_id', 'second_id'];
}


$t = new Test();
$t->first_id = 1;
$t->second_id = 2;
$t->save(); // <- Fails with telescope enabled
```

I had a lot of bug and exception starting to pop up after enabling telescope in a project where a couple of models require a composite key, and I had to disable telescope to solve it.

PR #1655, opened 2025-11-01
fix: skip watch models with composite primary key

artengin:
Refs: https://github.com/laravel/telescope/issues/1654


artengin:
Hi! I’ve made some changes so that models with composite primary keys are now safely ignored by Telescope. The previous error no longer occurs.
```php
array_key_exists(): Argument #1 ($key) must be a valid array offset type
```
I hope this contribution is helpful. Thank you!

taylorotwell:

PR #1656, opened 2025-11-10
Support MongoDb database

theofanisv:
Hello,

**I have added support to use MongoDB as a storage driver.**
I have been using Telescope in production for a few years now, and for some deployments, MySQL collects ~6 millions of records each day. After a few days, querying is degraded. MongoDB is supposed to handle that workload better, especially for `telescope_entries_tags`. 
Furthermore, we can [MongoDB's TTL](https://www.mongodb.com/docs/manual/tutorial/expire-data/) to expire old records, so we can omit the PruneCommand running as a cron job. This feature must be set up manually by the developer, it is not automatically handled by this update.

Also, I optimized the `MongoDbEntriesRepository` so that during the ClearCommand, to truncate the tables in a single command instead of deleting in batches for better performance.

I should note I have not yet used this version in production. If you would prefer it to mature, I can reopen this PR in a few months. 

To use MongoDB for Telescope, it is considered a prerequisite to [install MongoDB for Laravel](https://laravel.com/docs/mongodb) even though I have not noted it anywhere specifically.

I should note that even though I added `MongoDbDatabaseEntriesRepositoryTest`, I did not manage to make it run due to a weird error. Take that into consideration. 

Any comments are welcome.


taylorotwell:

PR #1667, opened 2025-12-15
Change Verb Colors

JonPurvis:
Hey 👋 

Tools like Postman and Swagger often assign different colours for each HTTP verb to make them more distinguishable. There doesn't actually seem to be a standard for this, so I've taken what Postman does and applied it to Telescope, so this:

Screenshot 2025-12-15 at 21 52 40

becomes:

Screenshot 2025-12-15 at 21 43 57

This makes it easier to tell apart a `POST` from a `PUT` and from a `PATCH`.


taylorotwell:
I think it's ok for now.

PR #1668, opened 2025-12-17
feat: add search query input for queries tab

haminh7036:
This PR add the search query input to the Queries tab in the RelatedEntries component.

**Benefit to end users:**
Users can find query they need to check quickly.

**How it makes building web applications easier:**
When debugging in the request, command, etc. has many queries (example over 100 queries). Scroll to find query is inconvenience. Developers can quickly find the query they need to check with search input.

**Screenshot:**
image
image


taylorotwell:

PR #1673, opened 2026-01-10
Fix PHPStan error in migration `getConnection`

osbre:
As Larastan/PHPStan becomes more popular, I think it's nice to comply with it "out of the box", so that users won't have to fix the error themselves.

`telescope:install` publishes this migration by default.


osbre:
Ah, nevermind. Turns out this package still preserves compatibility with Laravel <11.

PR #1678, opened 2026-01-30
hotfix: prevent duplicate migration creation during telescope:install.

keshavmansure:
**What does this PR do?**
This pull request fixes the issue where Telescope migrations were getting duplicated. The change ensures that migrations are created and executed only once, preventing conflicts and repeated entries.

**Problem**
Previously, Telescope migration files could be generated or run multiple times, leading to duplicate records and migration conflicts.

**Solution**
- Added proper checks to prevent duplicate Telescope migrations
- Ensured migration runs only when required
- Improved overall migration stability

**Impact**
- Prevents migration conflicts
- Keeps the database clean and consistent
- Safe change with no breaking impact on existing data

**🧪 Testing**
- Verified migrations run successfully without duplication
- Tested on a fresh setup and existing database

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -38,7 +38,7 @@ class InstallCommand extends Command
         $this->callSilent('vendor:publish', ['--tag' => 'telescope-config']);
 
         $this->comment('Publishing Telescope Migrations...');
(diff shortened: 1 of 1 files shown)


taylorotwell:

Thanks!

Issue #816, opened 2020-01-27
How can I custom auth guard in Telescope?


zhouzishu:
I have different guards in my Laravel project: `web` , `admin`, `api.` . But I can't use authorization for Telescope except `web` guard.

I think it'd better work with custom guard in config `telescope`, just like laravel nova, right?

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

danielpetrica:
Hi @zhouzishu did you get an answer somewhere else? I need the same guard change for one of my project but can't find anything else on google. Thank you and sorry for resurrecting this issue after so much time.

PR #1692, opened 2026-03-08
Add authentication guard configuration for Telescope dashboard

mahmoudmohamedramadan:
While working on **Telescope dashboard authorization**, I needed to use an `admin` guard instead of the default `web` guard.

During research, I found issue **#816**, but it was only a question, and there was no built-in way to configure the guard used by the `viewTelescope` Gate.

This PR addresses that gap by allowing developers to **specify the authentication guard** used when authorizing access to the Telescope dashboard.

### Configuration

A new `guard` option is added to the Telescope configuration:

```php
config/telescope.php

return [

    ...

    'guard' => 'web',

    ...

];
```

### Example (default `web` guard)

```php
/**
 * Register the Telescope gate.
 *
 * This gate determines who can access Telescope in non-local environments.
 */
protected function gate()
{
    Gate::define('viewTelescope', function ($user) {
        return in_array($user->email, [
            'user@guard.test',
        ]);
    });
}
```

### Example (using an `admin` guard)

```php
config/telescope.php

return [

    ...

    'guard' => 'admin',

    ...

];
```

```php
/**
 * Register the Telescope gate.
 *
 * This gate determines who can access Telescope in non-local environments.
 */
protected function gate()
{
    Gate::define('viewTelescope', function ($user) {
        return in_array($user->email, [
            'admin@guard.test',
        ]);
    });
}
```

This makes Telescope easier to integrate into applications that use **multiple guards** or **admin-specific authentication**.

### DIFF
--- a/config/telescope.php
+++ b/config/telescope.php
@@ -99,6 +99,19 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Telescope Authentication Guard
+    |--------------------------------------------------------------------------
+    |
+    | This option defines the authentication guard used to authorize access
+    | to the Telescope dashboard. You may change this value if your
+    | application uses a guard other than the default "web".
+    |
+    */
+
+    'guard' => 'web',
+
+    /*
+    |--------------------------------------------------------------------------
     | Allowed / Ignored Paths & Commands
     |--------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #1694, opened 2026-03-10
Support event-aware tags on queued listeners

rognales:
## Summary
- support queued listener `tags($event)` methods when Telescope extracts tags from `CallQueuedListener` jobs
- preserve existing zero-argument `tags()` behavior for listeners and jobs
- add regression coverage for both event-aware and zero-argument listener tags

## Problem
Horizon documents queued listener tagging with an event-aware signature, for example `tags(OrderShipped $event): array`.

Telescope currently extracts tags from queued listeners by instantiating the listener and calling `tags()` with no arguments, which raises an argument count error for that documented Horizon pattern.

## Approach
This change keeps the fix scoped to `ExtractTags::tagsForListener()`:
- pass the extracted event when the listener `tags` method accepts one parameter
- continue calling `tags()` with no arguments when the listener method requires none
- continue extracting tags from the queued event payload as before

## Testing
- `vendor/bin/phpunit tests/Telescope/ExtractTagTest.php`
- `vendor/bin/phpunit tests/Watchers/JobWatcherTest.php`

### DIFF
--- a/src/ExtractTags.php
+++ b/src/ExtractTags.php
@@ -9,6 +9,7 @@ use Illuminate\Events\CallQueuedListener;
 use Illuminate\Mail\SendQueuedMailable;
 use Illuminate\Notifications\SendQueuedNotifications;
 use ReflectionClass;
+use ReflectionMethod;
 use stdClass;
 
 class ExtractTags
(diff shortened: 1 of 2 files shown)


rognales:
Superseded by #1695.

PR #1695, opened 2026-03-10
Support queued listener tags that accept the event instance

### LINKED ISSUE #1693: Failed to extract tags for event listener
### Telescope Version

5.18.0

### Laravel Version

12.46.0

### PHP Version

8.4.18

### Database Driver & Version

MySQL 8.4.2 on macOS 14.8 arm64 (Homebrew client: 8.4.8)

### Description

Telescope appears to be incompatible with Laravel Horizon's documented listener tagging signature for queued event listeners.

Horizon documents listener tagging like this:
https://laravel.com/docs/12.x/horizon#manually-tagging-event-listeners

Example:

```php
use App\Events\OrderShipped;

class SendShipmentNotification implements ShouldQueue
{
    public function handle(OrderShipped $event): void
    {
        // ...
    }

    public function tags(OrderShipped $event): array
    {
        return ['shipment'];
    }
}
```

That works for Horizon, but when Telescope inspects a queued listener it calls `tags()` with no arguments and throws an argument count error.

From `Laravel\Telescope\ExtractTags`, Telescope ends up doing:

```php
return method_exists($target, 'tags') ? $target->tags() : [];
```
So for a queued listener with `public function tags(SomeEvent $event): array`, Telescope crashes with:

```console
Too few arguments to function App\Listeners\...\tags(), 0 passed in vendor/laravel/telescope/src/ExtractTags.php on line 102 and exactly 1 expected
````

### Expected Behavior
Telescope should not fail when inspecting a queued listener that uses Horizon's documented `tags()` signature.

### Steps

1. Install Laravel with Horizon and Telescope enabled.
2. Create a queued event listener implementing ShouldQueue.
3. Add a `tags(SomeEvent $event): array` method exactly as documented in the Horizon docs.
4. Dispatch the event so the queued listener runs. (don't forget to run horizon)
5. Let Telescope record the queue activity.

Example listener:

```php
use App\Events\OrderShipped;

class SendShipmentNotification implements ShouldQueue
{
    public function handle(OrderShipped $event): void
    {
        // ...
    }

    public function tags(OrderShipped $event): array
    {
        return ['shipment'];
    }
}
```

### Additional Context
This is specifically about queued listeners, not regular queued jobs.

It looks like `ExtractTags::tagsForListener()` inspects both:

- the listener instance created via reflection
- the queued event payload

but `explicitTags()` calls `tags()` without arguments on each target, which breaks for listener methods that expect the event.

rognales:
Closes #1693

## Summary
- support queued listener `tags($event)` methods when Telescope extracts tags from `CallQueuedListener` jobs
- preserve existing zero-argument `tags()` behavior for listeners and jobs
- add regression coverage for both event-aware and zero-argument listener tags

## Problem
Laravel Horizon documents queued listener tagging with an event-aware signature such as `tags(OrderShipped $event): array`.

Telescope currently instantiates queued listeners during tag extraction and calls `tags()` with no arguments. When an application follows the Horizon documentation for queued listener tags, Telescope throws an argument count error instead of recording the queue entry.

## Benefit To End Users
Applications using Horizon-style queued listener tagging can run Telescope without needing listener-specific workarounds such as nullable event parameters or duplicate tag extraction logic. This makes Telescope and Horizon behave consistently for the same queued listener pattern.

## Why This Does Not Break Existing Behavior
The change is scoped to queued listener tag extraction only.

- Listeners with `tags()` and no parameters continue to work as before.
- Jobs and other taggable objects still use the existing zero-argument `tags()` behavior.
- Queued listeners that accept the event instance in `tags($event)` now receive the extracted event payload.
- If no event object is present, Telescope still returns no explicit listener tags instead of failing.

## Implementation Details
`ExtractTags::tagsForListener()` now resolves explicit listener tags separately from generic target tag extraction:
- call `tags()` when the listener method requires no parameters
- call `tags($event)` when the listener method accepts a single parameter and the queued event object is available
- continue merging those tags with model-derived tags from the listener and tags extracted from the queued event itself

## Tests
- `vendor/bin/phpunit tests/Telescope/ExtractTagTest.php`
- `vendor/bin/phpunit tests/Watchers/JobWatcherTest.php`

The added regression coverage verifies both:
- queued listeners with `tags($event)`
- queued listeners with `tags()`

### DIFF
--- a/src/ExtractTags.php
+++ b/src/ExtractTags.php
@@ -9,6 +9,7 @@ use Illuminate\Events\CallQueuedListener;
 use Illuminate\Mail\SendQueuedMailable;
 use Illuminate\Notifications\SendQueuedNotifications;
 use ReflectionClass;
+use ReflectionMethod;
 use stdClass;
 
 class ExtractTags
@@ -83,11 +84,16 @@ class ExtractTags
      */
     protected static function tagsForListener($job)
     {
-        return collect(
-            [static::extractListener($job), static::extractEvent($job)]
-        )->map(function ($job) {
-            return static::from($job);
-        })->collapse()->unique()->toArray();
+        $listener = static::extractListener($job);
+        $event = static::extractEvent($job);
+
+        return collect([
+            static::explicitTagsForListener($listener, $event),
+            static::modelsFor([$listener])->map(function ($model) {
+                return FormatModel::given($model);
+            })->all(),
+            static::from($event),
+        ])->collapse()->unique()->toArray();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:

Thanks!

PR #1696, opened 2026-03-20
Fix Sentinel blocking Telescope behind Docker/reverse proxy

### LINKED ISSUE #1691: Local Telescope access breaks behind trusted local reverse proxy after Sentinel was added
### Description

Hi,

  After upgrading to a Telescope version that now depends on `laravel/sentinel`, Telescope stopped working in our local environment.

  Before Sentinel was added as a dependency, Telescope was working locally without issues.

 setup:
  - Laravel 12
  - Telescope 5.18.0
  - Docker
  - local environment
  - app is accessed through a local reverse proxy (nginx)
  - Laravel is configured to trust forwarded proxy headers because the same app is also deployed behind AWS ALB in non-local environments

  What happens now:
  - `/telescope` returns `401 Unauthorized` locally
  - this happens before Telescope’s normal authorization logic
  - the request is blocked by Sentinel middleware

  From debugging, the relevant part seems to be:
  - Telescope now prepends `Laravel\Sentinel\Http\Middleware\SentinelMiddleware`
  - Sentinel’s local driver rejects requests when Laravel considers the request to be from a trusted proxy and the resolved client IP is not private

  That means a local setup with trusted proxies can no longer access Telescope, even though this worked before Sentinel was introduced.
  if we remove/bypass Sentinel locally, Telescope works again

  Questions:
  1. Is this expected behavior?
  2. Is there an officially recommended way to keep trusted proxies enabled for deployed environments while still allowing normal local Telescope access?
  3. Would it make sense for Telescope to provide a config option to disable Sentinel protection in `local`, or to make the Sentinel layer opt-in?

  Thanks

JoshSalway:
Fixes #1691

## Problem

After Sentinel was added in v5.18.0, Telescope returns 401 when accessed locally through Docker/reverse proxy setups with trusted proxies configured. This affects Laravel Sail, Docker Desktop, and any local reverse proxy (nginx, Caddy, etc.).

**Root cause:** Sentinel's default Laravel driver has two issues:

1. `authorizeAccessingViaReverseProxies()` checks `$request->ip()` (resolved via `X-Forwarded-For` after trusted proxy headers), which can be a non-private IP after NAT translation — even though the actual connecting machine (`REMOTE_ADDR`) is on a private Docker network.

2. The tunnel hostname check (Expose, ngrok) is gated behind `! $request->isFromTrustedProxy()`, so configuring `TRUSTED_PROXIES=*` (standard in Sail/Docker) silently disables tunnel blocking entirely.

## Solution

Registers a custom `telescope` Sentinel driver via `Sentinel::extend()` that fixes both issues by changing the check order:

1. **Non-local env** → pass through (same as default driver)
2. **Always check tunnel hostnames** (`.sharedwithexpose.com`, `.ngrok-free.app`, `.ngrok.io`) → block, **regardless of proxy configuration**
3. **Check raw `REMOTE_ADDR`** (before trusted proxy resolution) → if private, allow (Docker/local proxy)
4. **Fallback** to parent's `authorizeAccessingViaReverseProxies()` for edge cases

### Why this addresses the Expose concern

[As noted in the previous PR](https://github.com/JoshSalway/telescope/pull/1#discussion_r2006662651), Expose sets `REMOTE_ADDR` to `127.0.0.1` even when accessed from a different device/network. Simply checking if `REMOTE_ADDR` is private would defeat Sentinel's security.

This driver solves it by checking the **hostname first**: tunnel hostnames are blocked before the private-IP check ever runs. So Expose (`REMOTE_ADDR=127.0.0.1`, `Host: *.sharedwithexpose.com`) is correctly blocked, while Docker (`REMOTE_ADDR=172.18.0.1`, `Host: localhost`) is correctly allowed.

### Unit tests (25 tests, 25 assertions — all pass)

Covers: normal local dev, IPv6 loopback, Docker proxy (172.x, 10.x, 192.168.x, wildcard), Herd `.test` domains, Expose tunnel (with/without proxy, with private forwarded IP), ngrok tunnel (free/io, with proxy), public IP edge cases, tunnels in production, driver resolution through Sentinel manager, and full middleware pipeline integration.

### Live HTTP tests against a real Laravel 12 app (`APP_ENV=local`, `TRUSTED_PROXIES=*`)

| Scenario | Host | X-Forwarded-For | Expected | Actual |
|----------|------|-----------------|----------|--------|
| Normal localhost | `localhost` | — | 200 | ✅ 200 |
| Herd `.test` domain | `myapp.test` | — | 200 | ✅ 200 |
| Docker behind proxy | `localhost` | `172.67.152.165` | 200 | ✅ 200 |
| Expose tunnel | `*.sharedwithexpose.com` | — | 401 | ✅ 401 |
| Expose + trusted proxy | `*.sharedwithexpose.com` | `203.0.113.50` | 401 | ✅ 401 |
| Expose + private fwd IP | `*.sharedwithexpose.com` | `192.168.1.100` | 401 | ✅ 401 |
| ngrok-free tunnel | `*.ngrok-free.app` | — | 401 | ✅ 401 |
| ngrok + trusted proxy | `*.ngrok-free.app` | `203.0.113.50` | 401 | ✅ 401 |
| ngrok.io tunnel | `*.ngrok.io` | — | 401 | ✅ 401 |

## Files changed

- **`src/TelescopeSentinelDriver.php`** (new) — Custom Sentinel driver
- **`src/TelescopeServiceProvider.php`** (+3 lines) — Register the driver
- **`tests/TelescopeSentinelDriverTest.php`** (new) — 25 tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/TelescopeSentinelDriver.php
+++ b/src/TelescopeSentinelDriver.php
@@ -1,0 +1,51 @@
+<?php
+
+namespace Laravel\Telescope;
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Str;
+use Laravel\Sentinel\Drivers\Driver;
+
+class TelescopeSentinelDriver extends Driver
+{
+    /**
+     * The tunnel service hostnames that should be blocked.
+     *
+     * @var array<int, string>
+     */
+    protected $tunnelHostnames = [
+        '.sharedwithexpose.com',
+        '.ngrok-free.app',
+        '.ngrok.io',
+    ];
+
+    /**
+     * Authorize access for the request.
+     */
+    public function authorize(Request $request): bool
+    {
+        if (! $this->app()->environment('local')) {
+            return true;
+        }
+
+        if ($this->isTunnelRequest($request)) {
+            return false;
+        }
+
+        $remoteAddr = $request->server->get('REMOTE_ADDR', '');
+
+        if ($remoteAddr !== '' && $this->isPrivateIp($remoteAddr)) {
+            return true;
+        }
+
+        return $this->authorizeAccessingViaReverseProxies($request);
+    }
+
+    /**
+     * Determine if the request is coming through a known tunnel service.
+     */
+    protected function isTunnelRequest(Request $request): bool
+    {
+        return Str::endsWith($request->host(), $this->tunnelHostnames);
+    }
+}
--- a/src/TelescopeServiceProvider.php
+++ b/src/TelescopeServiceProvider.php
@@ -5,6 +5,7 @@ namespace Laravel\Telescope;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\ServiceProvider;
 use Laravel\Sentinel\Http\Middleware\SentinelMiddleware;
+use Laravel\Sentinel\Sentinel;
 use Laravel\Telescope\Actions\UninstallAction;
 use Laravel\Telescope\Contracts\ClearableRepository;
 use Laravel\Telescope\Contracts\EntriesRepository;
@@ -27,6 +28,8 @@ class TelescopeServiceProvider extends ServiceProvider
         if (! config('telescope.enabled')) {
             return;
         }
+
+        Sentinel::extend('telescope', fn ($app) => new TelescopeSentinelDriver(fn () => $app));
 
         Route::middlewareGroup('telescope', [
             SentinelMiddleware::class.':telescope',
--- a/tests/TelescopeSentinelDriverTest.php
+++ b/tests/TelescopeSentinelDriverTest.php
@@ -1,0 +1,379 @@
+<?php
+
+namespace Laravel\Telescope\Tests;
+
+use Illuminate\Http\Request;
+use Laravel\Sentinel\Sentinel;
+use Laravel\Telescope\TelescopeSentinelDriver;
+use Symfony\Component\HttpFoundation\Request as SymfonyRequest;
+
+class TelescopeSentinelDriverTest extends FeatureTestCase
+{
+    /**
+     * Create a driver instance with the given environment.
+     */
+    protected function createDriver(string $environment = 'local'): TelescopeSentinelDriver
+    {
+        return new TelescopeSentinelDriver(function () use ($environment) {
+            $app = $this->app;
+            $app->detectEnvironment(fn () => $environment);
+
+            return $app;
+        });
+    }
+
+    /**
+     * Create a request with the given parameters.
+     */
+    protected function createRequest(
+        string $remoteAddr = '127.0.0.1',
+        string $host = 'localhost',
+        array $trustedProxies = [],
+        ?string $forwardedFor = null,
+    ): Request {
+        $server = [
+            'REMOTE_ADDR' => $remoteAddr,
+            'HTTP_HOST' => $host,
+            'SERVER_NAME' => $host,
+        ];
+
+        if ($forwardedFor !== null) {
+            $server['HTTP_X_FORWARDED_FOR'] = $forwardedFor;
+        }
+
+        $symfonyRequest = new SymfonyRequest(
+            query: [],
+            request: [],
+            attributes: [],
+            cookies: [],
+            files: [],
+            server: $server,
+        );
+
+        if (! empty($trustedProxies)) {
+            Request::setTrustedProxies($trustedProxies, Request::HEADER_X_FORWARDED_FOR);
+        }
+
+        return Request::createFromBase($symfonyRequest);
+    }
+
+    /** {@inheritdoc} */
+    #[\Override]
+    protected function tearDown(): void
+    {
+        Request::setTrustedProxies([], Request::HEADER_X_FORWARDED_FOR);
+
+        parent::tearDown();
+    }
+
+    // ---------------------------------------------------------------
+    // Non-local environment — Sentinel should never block
+    // ---------------------------------------------------------------
+
+    public function test_non_local_environment_always_allows_access(): void
+    {
+        $driver = $this->createDriver('production');
+
+        $request = $this->createRequest('203.0.113.50', 'myapp.com');
+        $this->assertTrue($driver->authorize($request));
+    }
+
+    public function test_staging_environment_always_allows_access(): void
+    {
+        $driver = $this->createDriver('staging');
+
+        $request = $this->createRequest('203.0.113.50', 'staging.myapp.com');
+        $this->assertTrue($driver->authorize($request));
+    }
+
+    // ---------------------------------------------------------------
+    // Normal local development — should allow
+    // ---------------------------------------------------------------
+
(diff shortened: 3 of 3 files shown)


JoshSalway:
## How to test locally

### 1. Set up a test app with this branch

```bash
laravel new telescope-sentinel-test
cd telescope-sentinel-test
composer config repositories.telescope vcs https://github.com/JoshSalway/telescope
composer require laravel/telescope:dev-fix/sentinel-docker-proxy-tunnel-blocking -W
php artisan telescope:install
php artisan migrate
```

Add `TRUSTED_PROXIES=*` to `.env` (simulates Docker/Sail setup). Confirm `APP_ENV=local`.

### 2. Start the server

```bash
php artisan serve --port=8099
```

### 3. Run the test scenarios

```bash
# Normal localhost (should return 200)
curl -s -o /dev/null -w "%{http_code}" http://localhost:8099/telescope

# Herd .test domain (should return 200)
curl -s -o /dev/null -w "%{http_code}" -H "Host: myapp.test" http://localhost:8099/telescope

# Docker proxy — the #1691 fix (should return 200)
curl -s -o /dev/null -w "%{http_code}" -H "X-Forwarded-For: 172.67.152.165" http://localhost:8099/telescope

# Expose tunnel (should return 401)
curl -s -o /dev/null -w "%{http_code}" -H "Host: myapp.sharedwithexpose.com" http://localhost:8099/telescope

# Expose tunnel + trusted proxy (should return 401)
curl -s -o /dev/null -w "%{http_code}" -H "Host: myapp.sharedwithexpose.com" -H "X-Forwarded-For: 203.0.113.50" http://localhost:8099/telescope

# Expose tunnel + private forwarded IP (should return 401)
curl -s -o /dev/null -w "%{http_code}" -H "Host: myapp.sharedwithexpose.com" -H "X-Forwarded-For: 192.168.1.100" http://localhost:8099/telescope

# ngrok-free tunnel (should return 401)
curl -s -o /dev/null -w "%{http_code}" -H "Host: abc123.ngrok-free.app" http://localhost:8099/telescope

# ngrok tunnel + trusted proxy (should return 401)
curl -s -o /dev/null -w "%{http_code}" -H "Host: abc123.ngrok-free.app" -H "X-Forwarded-For: 203.0.113.50" http://localhost:8099/telescope

# ngrok.io tunnel (should return 401)
curl -s -o /dev/null -w "%{http_code}" -H "Host: abc123.ngrok.io" http://localhost:8099/telescope
```

### 4. If you have Herd Pro with Expose

Share the test site via Herd's "Share" button, then visit the Expose URL (e.g. `https://telescope-sentinel-test.sharedwithexpose.com/telescope`). It should return **401**.

### 5. Run the unit tests

```bash
cd vendor/laravel/telescope
../../bin/phpunit tests/TelescopeSentinelDriverTest.php --testdox
```

All 25 tests should pass.

### Expected results

| # | Scenario | Expected |
|---|----------|----------|
| 1 | Normal localhost | 200 |
| 2 | Herd `.test` domain | 200 |
| 3 | Docker proxy (`X-Forwarded-For` with non-private IP) | 200 |
| 4 | Expose tunnel | 401 |
| 5 | Expose tunnel + trusted proxy | 401 |
| 6 | Expose tunnel + private forwarded IP | 401 |
| 7 | ngrok-free tunnel | 401 |
| 8 | ngrok + trusted proxy | 401 |
| 9 | ngrok.io tunnel | 401 |

JoshSalway:
## Breaking changes & versioning

This is **not a breaking change**:

- No public API changes (no method signatures changed, no config keys added/removed)
- No new dependencies (already requires `laravel/sentinel: ^1.0`)
- No user action required on upgrade

This is suitable for a **patch release** (e.g. `v5.18.1`).

### Behavioral change to be aware of

There is one intentional behavioral change: users who had `TRUSTED_PROXIES` configured and were accessing Telescope through Expose/ngrok tunnels would have been **accidentally allowed** through (due to the tunnel check being bypassed when trusted proxies are configured). After this fix, those tunnel requests will be correctly **blocked with 401**.

This is fixing a security gap, not breaking intended functionality — but worth noting for the changelog.

## Known limitations

These are inherited from Sentinel's hostname-based approach and are **not regressions** introduced by this PR:

1. **Unknown tunnel services** — Only Expose (`.sharedwithexpose.com`), ngrok-free (`.ngrok-free.app`), and ngrok (`.ngrok.io`) are detected. Other tunnel services (Cloudflare Tunnel, localhost.run, bore.pub, etc.) would not be caught if their `REMOTE_ADDR` is private. This matches the same hostname list used by Sentinel's default Laravel driver.

2. **Custom tunnel domains** — ngrok paid plans allow custom domains (e.g. `tunnel.mycompany.com`). These would bypass hostname detection. Again, same limitation as Sentinel itself.

3. **Host header spoofing** — Tunnel detection relies on `$request->host()` which reads the HTTP `Host` header. An attacker on the local network could theoretically set `Host: localhost` while routing through an unrecognized tunnel. However, this requires the attacker to already have access to the private network, which largely defeats the purpose of the attack.

4. **Hostname list is not user-configurable** — The blocked tunnel hostnames are hardcoded in the driver (matching Sentinel's approach). If configurability is desired, the `$tunnelHostnames` property is `protected` so it could be extended, or a config key could be added in a future iteration.

### Scope note

This fix is intentionally scoped to Telescope. If the consensus is that this logic belongs in `laravel/sentinel` itself (so all packages using Sentinel benefit), this driver serves as a working reference implementation for that change. The two approaches are not mutually exclusive — if Sentinel adopts this logic natively, the custom driver registration in Telescope could simply be removed.

src/TelescopeSentinelDriver.php
```diff
@@ -0,0 +1,51 @@
...
+    /**
+     * Authorize access for the request.
+     */
+    public function authorize(Request $request): bool
+    {
+        if (! $this->app()->environment('local')) {
+            return true;
+        }
+
+        if ($this->isTunnelRequest($request)) {
```
crynobone, line 31: This doesn't handle any other reverse proxies. Cloudflare Tunnel for example allows custom domain name. 

If you want to exclude Docker network, you need to PR a solution that cater for Docker.

crynobone (changes requested):

src/TelescopeSentinelDriver.php
crynobone: > Unknown tunnel services — Only Expose (.sharedwithexpose.com), ngrok-free (.ngrok-free.app), and ngrok (.ngrok.io) are detected. Other tunnel services (Cloudflare Tunnel, localhost.run, bore.pub, etc.) would not be caught if their REMOTE_ADDR is private. This matches the same hostname list used by Sentinel's default Laravel driver.

You own AI generated description explained this, and why this can't be merged.

JoshSalway:
Closing in favour of a new PR with a configurable allowlist approach instead of hostname blocklist, per @crynobone's feedback.

PR #1726, opened 2026-05-21
Allow filter by tags with logic AND

christhofer:
Currently, Telescope supports filter result by tag with comma separated using logic OR
```
/telescope/requests?tag=Auth%3A1,status%3A422
```
This will show results with `Auth:1` OR `status:422`

This Pull Request allow us to filter result by tag using logic AND, using `tags` query.
```
/telescope/requests?tags=Auth%3A1,status%3A422
```
This will show results that have both `Auth:1` AND `status:422`

This feature has been [discussed](https://github.com/laravel/framework/discussions/36372) in the main repo.

Caveat:
- No UI to use this logic yet, user needs to manually type the query params
- I didn't push the app.js from `npm run build`, should I?

### DIFF
--- a/resources/js/components/IndexScreen.vue
+++ b/resources/js/components/IndexScreen.vue
@@ -15,6 +15,7 @@ export default {
     data() {
         return {
             tag: '',
+            tags: '',
             familyHash: '',
             entries: [],
             ready: false,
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #1729, opened 2026-06-03
Support formatting composite model keys

### LINKED ISSUE #1654: Crash saving a model with a composite key - Only when telescope is enabled

ibobgunardi:
Fixes #1654.

`FormatModel::given()` currently calls Eloquent's `getKey()` for non-pivot models. When a model defines an array primary key, that eventually passes an array into `getAttribute()` and Telescope crashes while recording the model event.

This keeps the existing pivot behavior and formats models with array key names by reading each key attribute directly, producing the same `Class:key_key` tag shape Telescope already uses for multi-part pivot identifiers.

Tests run:

- `php -l src/FormatModel.php`
- `php -l tests/Watchers/ModelWatcherTest.php`
- `php vendor/bin/phpunit tests/Watchers/ModelWatcherTest.php --filter composite`
- `php vendor/bin/phpstan analyse src/FormatModel.php --memory-limit=1G`

Local note: running the whole `tests/Watchers/ModelWatcherTest.php` file in my unlocked Windows dependency install still has the pre-existing query-entry ordering failures in the older assertions; the new composite-key regression passes.

### DIFF
--- a/src/FormatModel.php
+++ b/src/FormatModel.php
@@ -16,11 +16,17 @@ class FormatModel
      */
     public static function given($model)
     {
+        $keyName = $model->getKeyName();
+
         if ($model instanceof Pivot && ! $model->incrementing) {
             $keys = [
                 $model->getAttribute($model->getForeignKey()),
                 $model->getAttribute($model->getRelatedKey()),
             ];
+        } elseif (is_array($keyName)) {
+            $keys = array_map(function ($key) use ($model) {
+                return $model->getAttribute($key);
+            }, $keyName);
         } else {
             $keys = $model->getKey();
         }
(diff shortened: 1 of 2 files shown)


taylorotwell:
Eloquent doesn't support composite primary keys.

PR #1741, opened 2026-06-20
[5.x] Clear stale exception and failed tag when a job is marked processed

EhabAmawi:
Fixes #1740.

## Problem

When the same queued job (same `telescope_uuid`) emits **both** a `JobFailed` and a `JobProcessed` event, `JobWatcher` persists an internally inconsistent job entry: `status: processed` while still carrying the `exception` payload **and** the `failed` tag written by the earlier failure.

`recordProcessedJob()` only updated the status and never cleared a previously-recorded exception:

```php
Telescope::recordUpdate(EntryUpdate::make(
    $uuid, EntryType::JOB, ['status' => 'processed']   // exception from a prior recordFailedJob() is retained
));
```

Because `DatabaseEntriesRepository::update()` merges changes with `array_merge($existingContent, $update->changes)`, the `exception` key set by `recordFailedJob()` survives, and the `failed` tag (added via `->addTags(['failed'])`) is never removed.

### How one `telescope_uuid` gets both events

A long-running job is reserved by two workers. With `tries = 1` and a queue `retry_after` shorter than the job's runtime, the in-flight job becomes visible again; a second worker reserves it, `attempts (2) > maxTries (1)`, and `Worker::process()` throws `MaxAttemptsExceededException` **before** `handle()` runs → `JobFailed`. The original worker later finishes successfully → `JobProcessed`. When the `JobProcessed` update is flushed last, the entry ends as **processed + exception + `failed` tag**, and the UI visibly flips failed → processed.

## Fix

Clear the stale failure state when marking a job processed:

```php
Telescope::recordUpdate(EntryUpdate::make(
    $uuid, EntryType::JOB, ['status' => 'processed', 'exception' => null]
)->removeTags(['failed']));
```

`exception` is set to `null` (the `array_merge` storage path cannot unset a key), which keeps the entry consistent and leaves `IncomingEntry::isFailedJob()` — which checks `status === 'failed'` — unaffected.

## Test

Added `test_processed_job_clears_stale_failure_state_left_by_a_duplicate_reservation` to `JobWatcherTest`. It records a pending job, fires `JobFailed` then `JobProcessed` for the same `telescope_uuid`, and asserts the stored entry is `processed`, has no lingering `exception`, and is no longer tagged `failed`. The test fails on the previous behaviour and passes with this change.

### DIFF
--- a/src/Watchers/JobWatcher.php
+++ b/src/Watchers/JobWatcher.php
@@ -101,8 +101,8 @@ class JobWatcher extends Watcher
         }
 
         Telescope::recordUpdate(EntryUpdate::make(
-            $uuid, EntryType::JOB, ['status' => 'processed']
-        ));
+            $uuid, EntryType::JOB, ['status' => 'processed', 'exception' => null]
+        )->removeTags(['failed']));
 
         $this->updateBatch($event->job->payload());
     }
--- a/tests/Watchers/JobWatcherTest.php
+++ b/tests/Watchers/JobWatcherTest.php
@@ -7,13 +7,17 @@ use Illuminate\Contracts\Bus\Dispatcher;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Auth\User;
+use Illuminate\Queue\Events\JobFailed;
+use Illuminate\Queue\Events\JobProcessed;
 use Illuminate\Queue\Jobs\Job;
 use Illuminate\Queue\QueueManager;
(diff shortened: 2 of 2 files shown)


EhabAmawi:
Heads-up: the failing checks here look unrelated to this change.

`LogWatcherTest` is currently red on `5.x` itself — the scheduled run on the same base commit fails with the same 17 `Attempt to read property "type" on null` errors (`test_log_watcher_*`) across Laravel 9–13, while Laravel 8 (older framework) passes. This PR only touches `JobWatcher`, and `JobWatcherTest` passes on CI.

Happy to rebase / re-run once `5.x` is green again.

PR #1742, opened 2026-06-22
Skip publishing migrations if telescope migration already exists

### LINKED ISSUE #1589: telescope:install create duplicate migrations
Image

### Telescope Version

5.7.0

### Laravel Version

12.12.0

### PHP Version

8.4.6

### Database Driver & Version

10.4.32-MariaDB

### Description

when you simply run `php artisan telescope:install` for the first time in your project it will place a new migration file in your app, the problem if you ran the command one more time it will place the same file again but with a new name, causing a migration error `telescope_entries` already exists.

### Steps To Reproduce

1. in a new or existing project call `php artisan telescope:install`.
2. now run `php artisan migrate`, if this is a new project everything will go ok.
3. now call again `php artisan telescope:install`, a new migration file will be created inside app migrations folder with the different name but the same content from step no.1.

lazerg:
Running \`php artisan telescope:install\` more than once publishes a new migration file with a fresh timestamp each time, because \`publishesMigrations()\` (Laravel 11+) prefixes files with the current date. The second migration then fails with "table telescope_entries already exists."

The fix checks for any existing \`*_create_telescope_entries_table.php\` file in \`database/migrations/\` before publishing, and skips the publish step if one is already present — matching the same guard pattern used in the service provider registration below it.

Closes #1589.

### DIFF
--- a/src/Console/InstallCommand.php
+++ b/src/Console/InstallCommand.php
@@ -38,7 +38,10 @@ class InstallCommand extends Command
         $this->callSilent('vendor:publish', ['--tag' => 'telescope-config']);
 
         $this->comment('Publishing Telescope Migrations...');
-        $this->callSilent('vendor:publish', ['--tag' => 'telescope-migrations']);
+
+        if (! glob(database_path('migrations/*_create_telescope_entries_table.php'))) {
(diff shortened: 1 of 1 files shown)


lazerg:
Closing — the issue author already has an unmerged PR for this.

PR #1751, opened 2026-07-29
[5.x] Retry exception storage on deadlock to survive concurrent job failures

lazerg:
Exceptions are stored by flipping `should_display_on_index` on the existing rows of the same family and then inserting the new row. When several Horizon workers fail jobs at the same time, these concurrent update/insert pairs contend on the same index range and MySQL kills one of them with a deadlock (1213), which then surfaces out of the failed-job handler instead of just being recorded. Wrapping each chunk of exception writes in a transaction lets the framework retry them when that happens.

Fixes #1750


lazerg:
These CI failures already happen on 5.x itself. The latest scheduled run on the branch reports the same 17 errors and 23 failures, all in the request watcher tests, so they predate this change and aren't caused by it. The test added here passes on every version that can run it.

lazerg:
These CI failures aren't from me, they're already happening on 5.x.

taylorotwell:

PR #1754, opened 2026-08-07
Fix authorization failure when trustProxies is enabled

Tahazargar:
Problem

When an app trusts all proxies ($middleware->trustProxies(at: '*')),
request()->ip() resolves to the X-Forwarded-For header instead of
the real socket IP. Telescope’s default gate() uses this IP to check
for local access, so it incorrectly returns 401 even for local requests.

Solution

Fall back to $_SERVER['REMOTE_ADDR'] in the authorization check,
which always reflects the actual TCP peer IP and cannot be spoofed
via proxy headers.

### DIFF
--- a/src/TelescopeApplicationServiceProvider.php
+++ b/src/TelescopeApplicationServiceProvider.php
@@ -27,8 +27,9 @@ class TelescopeApplicationServiceProvider extends ServiceProvider
         $this->gate();
 
         Telescope::auth(function ($request) {
-            return app()->environment('local') ||
-                   Gate::check('viewTelescope', [$request->user()]);
+            return app()->environment('local')
+                || in_array($_SERVER['REMOTE_ADDR'] ?? null, ['127.0.0.1', '::1'])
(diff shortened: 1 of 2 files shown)


src/TelescopeApplicationServiceProvider.php
```diff
@@ -27,8 +27,9 @@ protected function authorization()
         $this->gate();
 
         Telescope::auth(function ($request) {
-            return app()->environment('local') ||
-                   Gate::check('viewTelescope', [$request->user()]);
+            return app()->environment('local')
+                || in_array($_SERVER['REMOTE_ADDR'] ?? null, ['127.0.0.1', '::1'])
```
crynobone, line 31: This may lead the security vulnerability if you use `local` environment and serve the app via reverse proxy such as ngork.

At this points, it isn't safe to rely on `REMOTE_ADDR` as any request using ngork will return `127.0.0.1`

crynobone (changes requested):

PR #1758, opened 2026-08-26
[5.x] Add XML payload and response support to ClientRequestWatcher

haminh7036:
This PR adds support for parsing XML request payloads and responses in `ClientRequestWatcher`.

Currently, when sending or receiving XML via the HTTP Client, Telescope doesn't parse the XML body—recording payloads as empty arrays and responses as `HTML Response`.

With this change, valid XML content is parsed into arrays so it can be inspected, masked (`hideParameters`), and displayed cleanly in the UI similar to JSON payloads.


taylorotwell:

PR #1765, opened 2026-09-08
[5.x] Fix SQL corruption in QueryWatcher when named bindings share a prefix

haminh7036:
When a query uses named placeholders where one is a prefix of another (e.g. `:content1` and `:content10`), `QueryWatcher::replaceBindings()` matches and replaces the prefix inside the longer placeholder. This leaves trailing characters (such as `'a'0`) and corrupts the resulting SQL.

### DIFF
--- a/src/Watchers/QueryWatcher.php
+++ b/src/Watchers/QueryWatcher.php
@@ -96,7 +96,7 @@ class QueryWatcher extends Watcher
         foreach ($this->formatBindings($event) as $key => $binding) {
             $regex = is_numeric($key)
                 ? "/\?(?=(?:[^'\\\']*'[^'\\\']*')*[^'\\\']*$)/"
-                : "/:{$key}(?=(?:[^'\\\']*'[^'\\\']*')*[^'\\\']*$)/";
+                : "/:{$key}(?![a-zA-Z0-9_])(?=(?:[^'\\\']*'[^'\\\']*')*[^'\\\']*$)/";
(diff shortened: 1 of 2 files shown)


taylorotwell:
Laravel doesn't use named place-holders like that?

haminh7036:
It happens in raw bulk queries with indexed placeholders (e.g. bulk replacing product attributes):

```php
DB::statement('
    REPLACE INTO product_attributes (product_code, column_id, column_value) VALUES 
    (:product_code1, :column_id1, :column_value1),
    (:product_code2, :column_id2, :column_value2),
    ...
    (:product_code10, :column_id10, :column_value10)
', [
    'product_code1' => 'P01', 'column_id1' => 1, 'column_value1' => 'Red',
    'product_code2' => 'P01', 'column_id2' => 2, 'column_value2' => 'XL',
    // ...
    'product_code10' => 'P01', 'column_id10' => 10, 'column_value10' => 'Cotton',
]);
```

Because Telescope already supports named bindings, `:column_value1` matches the prefix of `:column_value10`, corrupting it into `'Red'0` in the logged SQL.

This PR adds `(?![a-zA-Z0-9_])` so the regex only matches the full placeholder name.

haminh7036:
@taylorotwell Just following up on this when you have a moment. I actually ran into this at work where we dynamically generate indexed named placeholders for bulk queries. Thanks!

PR #1766, opened 2026-09-09
Add full UUID option to telescope:list command

ylynfatt:
The `telescope:list` table currently shortens entry UUIDs to eight characters. While this keeps the table compact, you can't copy the displayed value into `telescope:show`, which requires the full UUID.

This PR adds a `--full-uuid` option to allow developers the ability to display complete entry UUIDs:

```bash
  php artisan telescope:list request --full-uuid
```

The existing compact output remains the default, and the JSON output is unchanged. A test has been added to cover both the shortened default and the new full UUID output.

Here is an example of the default output:

```bash
$ php artisan telescope:list request --before=631 --limit=1

  +----------+--------+-----------------+--------+----------+----------------+
  | UUID     | Method | URI             | Status | Duration | Created        |
  +----------+--------+-----------------+--------+----------+----------------+
  | a2b3c0b0 | GET    | /telescope-demo | 200    | 6ms      | 46 minutes ago |
  +----------+--------+-----------------+--------+----------+----------------+
  Showing 1 entry - Use --before=630 for next page
```

and an example Output with `--full-uuid`:

```bash
$ php artisan telescope:list request --before=631 --limit=1 --full-uuid

  +--------------------------------------+--------+-----------------+--------+----------+----------------+
  | UUID                                 | Method | URI             | Status | Duration | Created        |
  +--------------------------------------+--------+-----------------+--------+----------+----------------+
  | a2b3c0b0-61e2-45ae-baab-f438d02b0044 | GET    | /telescope-demo | 200    | 6ms      | 46 minutes ago |
  +--------------------------------------+--------+-----------------+--------+----------+----------------+
  Showing 1 entry - Use --before=630 for next page
```

I hope this helps other developers, as it wasn't obvious to me at first that I couldn't use the shortened UUID directly with the `telescope:show` command, and I had to use the `--json` flag to get the full UUID.

### DIFF
--- a/src/Console/ListCommand.php
+++ b/src/Console/ListCommand.php
@@ -28,6 +28,7 @@ class ListCommand extends Command
         {--family= : Filter by family hash}
         {--limit=20 : Max entries to show}
         {--before= : Pagination cursor (sequence ID)}
+        {--full-uuid : Display complete entry UUIDs}
         {--json : Output entries as JSON}';
 
     /**
@@ -112,7 +113,7 @@ class ListCommand extends Command
             EntryType::REQUEST => [
                 ['UUID', 'Method', 'URI', 'Status', 'Duration', 'Created'],
                 fn ($entry) => [
-                    $this->shortUuid($entry->id),
+                    $this->formatUuid($entry->id),
                     $this->colorMethod($entry->content['method'] ?? ''),
                     Str::limit($entry->content['uri'] ?? '', 40),
                     $this->colorStatus((int) ($entry->content['response_status'] ?? 0)),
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks for spotting this.

The flag left the same problem in place elsewhere: `telescope:show` prints shortened UUIDs too (batch header, related queries, related exceptions), so drilling down from a `show` still dead-ended. Instead of a flag per table, `telescope:show` now accepts a shortened UUID directly, git-style:

```bash
$ php artisan telescope:list request --limit=1

  | UUID     | Method | URI             | Status | Duration |
  | a2b3c0b0 | GET    | /telescope-demo | 200    | 6ms      |

$ php artisan telescope:show a2b3c0b0
```

pushpak1300 (approved):

ylynfatt:
Okay, thanks, and noted, @pushpak1300. It's probably better to allow the shortened UUID rather than add another option. 👍🏽

